PR #33, opened 2023-05-10
Unique app name

Luc45:
Tweaks the suggested app name when authenticating as a partner so that it is unique and can be created multiple times without renaming or deleting the previous one.

image


zhongruige (approved):
LGTM! Thanks for this change @Luc45

PR #34, opened 2023-05-15
Use 'raw' download URL.

barryhughes:
Hi! Minor thing:

```
1. Run `wget https://github.com/woocommerce/qit-cli/blob/trunk/qit`
```

This fails because it is the URL of a webpage. If we want to fetch the binary with Wget/Curl etc we probably need to use [`https://github.com/woocommerce/qit-cli/raw/trunk/qit`](https://github.com/woocommerce/qit-cli/raw/trunk/qit).

### Testing instructions

1. Try following the updated instructions to install globally using `wget`.
2. Try to run the `qit` binary. It should work!

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,7 +27,7 @@ You can install QIT in three different ways:
 
 ### Globally Using `wget`
 
-1. Run `wget https://github.com/woocommerce/qit-cli/blob/trunk/qit`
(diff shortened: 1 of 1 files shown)


Luc45:
Thank you!

PR #37, opened 2023-05-22
Update activation test snapshots

Luc45:
This PR updates the snapshots for Activation Tests following the PR that does not include errors from WooCommerce Core in activation test results when Woo is not the SUT itself.


zhongruige (approved):
Thanks for getting these snapshots updated @Luc45!

PR #38, opened 2023-05-23
Update CLI to work with enums nested in items

Luc45:
This PR updates the CLI to handle an enum nested in an items schema from the WordPress REST API.


MrJnrman (approved):
Works like a charm! Thanks @Luc45

PR #39, opened 2023-05-23
Mininum CLI version

Luc45:
Tweaks the logic for enforcing updates by changing "enforce_latest" to showing a update nudge if available, or enforcing an udpate if the current version is below the minimum.


src/src/ManagerSync.php
```diff
@@ -92,18 +92,18 @@ public function enforce_latest_version(): void {
 		}
 
 		$latest_version = $this->environment->get_cache()->get_manager_sync_data( 'latest_cli_version' );
-		$enforce_latest = $this->environment->get_cache()->get_manager_sync_data( 'enforce_latest' );
+		$minimum_cli_version = $this->environment->get_cache()->get_manager_sync_data( 'minimum_cli_version' );
 
 		if ( version_compare( $current_version, $latest_version, '<' ) ) {
-			if ( $enforce_latest ) {
+			$this->output->writeln( '<comment>There\'s a new version of the QIT CLI available!</comment>' );
+			$this->output->writeln( sprintf( "Current version: %s\n<info>Latest version: %s</info>\nIt's strongly recommended to update. Update instructions: https://woocommerce.github.io/qit-documentation/#/cli/getting-started?id=updating-qit</info>\n", $current_version, $latest_version ) );
```
zhongruige: I'm debating a bit on the wording here, could we upsell a bit more, something like:

```suggestion
			$this->output->writeln( sprintf( "Current version: %s\n<info>Latest version: %s</info>\nUpdate today to take advantage of the latest features! Update instructions: https://woocommerce.github.io/qit-documentation/#/cli/getting-started?id=updating-qit</info>\n", $current_version, $latest_version ) );
```

src/src/ManagerSync.php
```diff
@@ -92,18 +92,18 @@ public function enforce_latest_version(): void {
 		}
 
 		$latest_version = $this->environment->get_cache()->get_manager_sync_data( 'latest_cli_version' );
-		$enforce_latest = $this->environment->get_cache()->get_manager_sync_data( 'enforce_latest' );
+		$minimum_cli_version = $this->environment->get_cache()->get_manager_sync_data( 'minimum_cli_version' );
 
 		if ( version_compare( $current_version, $latest_version, '<' ) ) {
-			if ( $enforce_latest ) {
+			$this->output->writeln( '<comment>There\'s a new version of the QIT CLI available!</comment>' );
```
MrJnrman: Can we update the wording here to have a red background so it is more prominent among the other text in the terminal?

Luc45: Thanks for the suggestion. What do you think about these?

**Using a version below the minimum set by the Manager:**
image

**Using an outdated, but still supported version:**
image

Luc45:
As part of this PR, I also added a `sync` command in Dev mode to ease of testing.

src/src/ManagerSync.php
MrJnrman: Thanks for making the change! Users definitely won't miss this

MrJnrman (approved):
Thanks for making the changes @Luc45! Retested and everything works as expected 🚢

PR #40, opened 2023-05-24
Update Self Tests

MrJnrman:
This PR adds an api self test to ensure that the `new_product_editor` feature can be enabled correctly. It also updates the existing snapshots based on the changes in https://github.com/Automattic/compatibility-dashboard/pull/414

### DIFF
--- a/_tests/api/valid_features/woocommerce-product-feeds/woocommerce-product-feeds.php
+++ b/_tests/api/valid_features/woocommerce-product-feeds/woocommerce-product-feeds.php
@@ -4,7 +4,7 @@
  * Plugin name: API - Log that the  New Product Editor is enabled.
  */
 
- add_action( 'woocommerce_admin_get_feature_config', static function( $features ) {
+ add_filter( 'woocommerce_admin_get_feature_config', static function( $features ) {
 	if ( 
 		array_key_exists( 'new-product-management-experience', $features ) &&
 		$features['new-product-management-experience'] === true
--- a/_tests/api/valid_features/woocommerce-product-feeds/woocommerce-product-feeds.php
+++ b/_tests/api/valid_features/woocommerce-product-feeds/woocommerce-product-feeds.php
@@ -11,4 +11,6 @@
 	) {
 		trigger_error( "New Product Editor is enabled as expected." );
 	}
+
+	return $features;
 }, 99 );
--- a/_tests/api/valid_features/woocommerce-product-feeds/woocommerce-product-feeds.php
+++ b/_tests/api/valid_features/woocommerce-product-feeds/woocommerce-product-feeds.php
@@ -1,0 +1,15 @@
+<?php
+
+/*
+ * Plugin name: API - Trigger failure when New Product Editor is not enabled.
+ */
+
+ add_action( 'woocommerce_admin_get_feature_config', static function( $features ) {
+	$features_controller = wc_get_container()->get( 'Automattic\WooCommerce\Internal\Features\FeaturesController' );
+	if ( 
+		array_key_exists( 'new-product-management-experience', $features ) &&
+		$features['new-product-management-experience'] === true
+	) {
+		trigger_error( "New Product Editor is enabled as expected." );
+	}
+}, 99 );
--- a/_tests/api/valid_features/woocommerce-product-feeds/woocommerce-product-feeds.php
+++ b/_tests/api/valid_features/woocommerce-product-feeds/woocommerce-product-feeds.php
@@ -5,7 +5,6 @@
  */
 
  add_action( 'woocommerce_admin_get_feature_config', static function( $features ) {
-	$features_controller = wc_get_container()->get( 'Automattic\WooCommerce\Internal\Features\FeaturesController' );
 	if ( 
 		array_key_exists( 'new-product-management-experience', $features ) &&
 		$features['new-product-management-experience'] === true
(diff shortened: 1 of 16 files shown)


_tests/tests/__snapshots__/ApiTest__test_api_valid_features__1.php
```diff
@@ -0,0 +1,1416 @@
...
+                    "count": "3676",
+                    "message": "PHP Warning: in_array() expects parameter 2 to be array, null given in \\/var\\/www\\/html\\/wp-content\\/plugins\\/woocommerce\\/src\\/Admin\\/Features\\/Features.php on line 195"
+                },
+                {
+                    "count": "611",
+                    "message": "PHP Warning: Invalid argument supplied for foreach() in \\/var\\/www\\/html\\/wp-content\\/plugins\\/woocommerce\\/src\\/Admin\\/Features\\/Features.php on line 135"
+                },
+                {
+                    "count": "4251",
+                    "message": "PHP Notice: New Product Editor is enabled as expected. in \\/var\\/www\\/html\\/wp-content\\/plugins\\/woocommerce-product-feeds\\/woocommerce-product-feeds.php on line 13"
```
Luc45: Awesome, we can see that it's activated here. However, there are some other notices being thrown around, I wonder if it has something to do with the way we enabled the feature plugin in the SUT?

```json
                   "count": "4287",
                    "message": "PHP Warning: array_filter() expects parameter 1 to be array, null given in \\/var\\/www\\/html\\/wp-content\\/plugins\\/woocommerce\\/src\\/Internal\\/Admin\\/FeaturePlugin.php on line 211"
                },
                {
                    "count": "4287",
                    "message": "PHP Warning: array_keys() expects parameter 1 to be array, null given in \\/var\\/www\\/html\\/wp-content\\/plugins\\/woocommerce\\/src\\/Internal\\/Admin\\/FeaturePlugin.php on line 211"
                },
                {
                    "count": "3676",
                    "message": "PHP Warning: array_diff(): Expected parameter 1 to be an array, null given in \\/var\\/www\\/html\\/wp-content\\/plugins\\/woocommerce\\/src\\/Admin\\/Features\\/Features.php on line 184"
                },
                {
                    "count": "3676",
                    "message": "PHP Warning: array_values() expects parameter 1 to be array, null given in \\/var\\/www\\/html\\/wp-content\\/plugins\\/woocommerce\\/src\\/Admin\\/Features\\/Features.php on line 184"
                },
                {
                    "count": "3676",
                    "message": "PHP Warning: in_array() expects parameter 2 to be array, null given in \\/var\\/www\\/html\\/wp-content\\/plugins\\/woocommerce\\/src\\/Admin\\/Features\\/Features.php on line 195"
                },
                {
                    "count": "611",
                    "message": "PHP Warning: Invalid argument supplied for foreach() in \\/var\\/www\\/html\\/wp-content\\/plugins\\/woocommerce\\/src\\/Admin\\/Features\\/Features.php on line 135"
                },
```

_tests/tests/__snapshots__/E2eTest__test_e2e_no_op__1.php
```diff
@@ -805,32 +805,32 @@
                     },
                     {
                         "file": "new-product-editor\\/new-product-editor.spec.js",
-                        "status": "passed",
+                        "status": "failed",
```
Luc45: Any idea why the Product editor tests started to fail on E2E, even though we didn't tweaked the e2e test env settings?

_tests/api/valid_features/woocommerce-product-feeds/woocommerce-product-feeds.php
```diff
@@ -0,0 +1,14 @@
+<?php
+
+/*
+ * Plugin name: API - Log that the  New Product Editor is enabled.
+ */
+
+ add_action( 'woocommerce_admin_get_feature_config', static function( $features ) {
```
Luc45: This should be `add_filter`, returning `$features` at the end. That's why we are receiving the warnings on the test log

MrJnrman: Good catch!

_tests/tests/__snapshots__/E2eTest__test_e2e_no_op__1.php
MrJnrman: I think this was from a previous run when I was modifying the e2e tests. Let me update the snapshots and see if it persists.

MrJnrman: Turns out that wasn't the reason. I checked the [test itself](https://github.com/Automattic/compatibility-dashboard/blob/trunk/ci/e2e/tests/new-product-editor/new-product-editor.spec.js) and it looks like it should fail if that specific test isn't skipped.

I think we should skip the product editor tests in general

_tests/tests/__snapshots__/ApiTest__test_api_valid_features__1.php
MrJnrman: Turns out it was because I wasn't using `add_filter` and I wasn't returning the `$features` array.

_tests/tests/__snapshots__/E2eTest__test_e2e_no_op__1.php
MrJnrman: As per DM chat, it turned out that the the `ENABLE_NEW_PRODUCT_EDITOR` where affecting the e2e tests. Changing the name to `QIT_ENABLE_NEW_PRODUCT_EDITOR` did the trick

Luc45 (approved):
Thank you for these self tests, Jamel! We can merge this once the Manager counterpart is merged.

PR #43, opened 2023-06-08
Accept Zips without parent directory under some circumstances

### LINKED ISSUE #44: Invalid zip contents despite correct zip file
When I try to upload my zip file for testing with `qit run:security woocommerce-pixel-manager --zip=woocommerce-pixel-manager.zip` I get the error: 

`Invalid zip contents. Please make sure that the zip contains the plugin folder at the root level. Expected to find directory: woocommerce-pixel-manager`

The zip file is correctly formatted. It contains the `woocommerce-pixel-manager` folder. And it also works fine when I upload the file for a new release. It gets processed by the PBS without problems. 

This seems to be a bug. 

Let me know if you need more information.

Luc45:
This PR addresses a bug where certain zipped files were being rejected by QIT, despite being acceptable for uploading to WordPress. The following scenarios have been identified and resolved:

### 1. Accepting Zips without a Parent Directory:

When you upload a zip without a parent directory to WordPress, it will use the zip filename as the slug of the plugin, for instance: `example.zip` will create `wp-content/plugins/example` and extract the contents to that directory. In this PR, we now accept a zip without a parent directory, as long as the zip file name matches the SUT slug.

### 2. Handling Malformed Zips with Missing Parent Directory in Zip Index:

In the case of a malformed zip where the parent directory is not included in the zip index, we have implemented a new approach to detect the presence of the root directory. Previously, we searched for an exact match of a directory that matches the SUT slug in the root using the zip index. However, now we indirectly check the presence of the root directory by examining the contents of the plugin. We iterate through the files in the zip, and if any file is found inside the expected directory, we consider the parent directory to be present.

The previous code snippet for searching the parent directory in the zip index was as follows:

```
for ( $i = 0; $i < $zip->numFiles; $i ++ ) {
	if ( $zip->statIndex( $i )['name'] === $plugin_slug . '/' ) {
		$found_parent_directory = true;
	}
}
```

The new implementation for searching the parent directory in the contents of the plugin is as follows:

```
for ( $i = 0; $i < $zip->numFiles; $i ++ ) {
	if ( str_starts_with( strtolower( $zip->statIndex( $i )['name'] ), "$plugin_slug/" ) ) {
		$found_parent_directory = true;
	}
}
```

In addition to the above improvements, we have also added a check for the presence of a PHP file with the "Plugin Name" header. This header is the only plugin header required by WordPress.

_(Testing instructions will be added in the Manager repo)_

Closes https://github.com/woocommerce/qit-cli/issues/44


src/src/Commands/Partner/AddPartner.php
```diff
@@ -166,7 +166,7 @@ protected function get_authorize_url( OutputInterface $output ): string {
 		$separator = stripos( $base_url, '?' ) === false ? '?' : '&';
 
 		// Generate a unique app name based on the current timestamp to avoid collision.
-		$app_name = urlencode( 'Woo Quality Insights Toolkit CLI (QIT) ' . time() );
+		$app_name = rawurlencode( 'Woo Quality Insights Toolkit CLI (QIT) ' . time() );
```
Luc45, line 169: This solves a PHPCS warning that was introduced in some other PR

Luc45:
Closed in favor of https://github.com/woocommerce/qit-cli/pull/46

PR #46, opened 2023-06-10
Accept Zips without parent directory under some circumstances (2)

### LINKED ISSUE #44: Invalid zip contents despite correct zip file

Luc45:
This PR is an alternative to PR https://github.com/woocommerce/qit-cli/pull/43

Closes https://github.com/woocommerce/qit-cli/issues/44

This PR is a simpler approach. Now, we only check that the file is a valid zip on the CLI, but we leave the more complex zip content validation checks to be done only on the Manager.

Testing instructions can be found in the Manager countepart to this PR.


zhongruige (approved):
Everything tested great for me, thanks for the work on this @Luc45!

PR #48, opened 2023-06-23
Update self-tests to use RC instead of a fixed version

Luc45:
This PR updates the WooCommerce version used in self tests to be RC instead of a fixed version, so that we can catch issues with new versions faster with the self tests, as using a fixed version masked an issue with an E2E test with the latest release.


zhongruige (approved):
LGTM thanks for this fix @Luc45!

PR #49, opened 2023-06-23
Update snapshots

Luc45:


_tests/tests/__snapshots__/E2eTest__test_e2e_no_op__1.php
```diff
@@ -68,27 +68,27 @@
...
                                 },
                                 {
                                     "title": "can complete the product types section",
-                                    "status": "passed"
+                                    "status": "pending"
                                 },
                                 {
                                     "title": "can complete the business section",
-                                    "status": "passed"
+                                    "status": "pending"
```
Luc45, line 91: Tests are passing, the number of successes and pending are fine, but I don't know why these are being skipped now, maybe we could double-check with Solaris

Luc45: @woocommerce/solaris any idea why these tests are being skipped on 7.9.0-beta.1?

Luc45:
I've went ahead and merged this anyway along with the Manager counterpart, as the "pending" tests doesn't seem to impact the overall test coverage.

_tests/tests/__snapshots__/E2eTest__test_e2e_no_op__1.php
lanej0: Hey @Luc45 -- some of the onboarding tests have been marked "skip" for a while now because development was going pretty fast with the onboarding wizard (I think it's called "core profiler" now).  Instead of constantly chasing failures, we just opted to skip some of the OBW tests for now.

Luc45: Thanks for the context @lanej0! That's super helpful to know!

PR #51, opened 2023-06-27
Add self-tests to PHPStan ignoring vendor directory

### LINKED ISSUE #47: exclude vendor directory for PHPStan test
When running the PHPStan test, I get many file errors from my plugin's `/vendor/` directory. 

In my opinion, that directory should be excluded, or we should have the option to exclude it.

Luc45:
Closes https://github.com/woocommerce/qit-cli/issues/47

This PR tweaks PHPStan tests so that it discovers the symbols in the vendor directories, but don't analyse the code on vendor directory.

Self-tests were added as part of this PR, the expected behavior is:

- `Foo` and `Bar` are found by PHPStan (not flagged as non-existing)
- `SomeUnexistingClassThatPHPStanShouldFlag` is flagged as non-existing as expected
- Violations on `PHPStanShouldNotFlagThis` and `PHPStanShouldNotFlagThis2` in the vendor directories are not flagged as expected.

[Related issue on PHPStan](https://github.com/phpstan/phpstan/issues/3749#issuecomment-756240488)

### DIFF
--- a/_tests/phpstan/with_vendor/env.php
+++ b/_tests/phpstan/with_vendor/env.php
@@ -1,0 +1,2 @@
+<?php
+return [];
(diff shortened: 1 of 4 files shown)


zhongruige (approved):
Thanks for adding in these self tests @Luc45!

PR #53, opened 2023-07-04
Update self tests for 7.8.2

Luc45:
This PR updates the self-tests for WooCommerce 7.8.2, which introduces no changes.

It also updates the status for `--wait` used by self-tests to wait for the new `running` status.


Luc45:
I'll be merging this as it's a pretty straightforward change and it's needed for self-tests.

PR #59, opened 2023-07-12
Update gitignore with common files

zhongruige:
This PR updates the QIT CLI `.gitignore` file with common files from IDEs and systems to prevent accidental commits and merges of these files.

**Review notes**
* Take a look at the list and see if there's anything else we should add/remove.


Luc45 (approved):
Thanks for this, Greg!

PR #62, opened 2023-07-17
Update self tests snapshots

Luc45:


Luc45:
Closing this PR as the snapshot change is from flakiness that is caused by parallel API tests that cause a deadlock in the DB, which is to some degree expected, unless we disable parallelization in API tests.

PR #65, opened 2023-07-21
Update self-test snapshots for custom DB

Luc45:


MrJnrman (approved):
Thanks for the change @Luc45!

Luc45:
Re-running this against Prod one last time and then I'll merge

PR #66, opened 2023-07-26
Update self tests snapshots

zhongruige:
This PR updates self tests snapshots.

PR #67, opened 2023-08-03
Update self tests snapshots

zhongruige:

### DIFF
--- a/_tests/tests/__snapshots__/ApiTest__test_api_delete_products__1.php
+++ b/_tests/tests/__snapshots__/ApiTest__test_api_delete_products__1.php
@@ -1381,10 +1381,6 @@
         {
             "debug_log": [
                 {
-                    "count": "1",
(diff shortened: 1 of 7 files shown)


zhongruige:
Looks like we're seeing the same results with two of the e2e tests as on https://github.com/woocommerce/qit-cli/pull/66

Luc45:
This is weird, we can see it's passing on Woo: https://github.com/woocommerce/woocommerce/actions/runs/5752528579/job/15593734491#step:7:234

I'll run a single E2E test against Staging to get the Allure Test Report and see where it's failing exactly.

Luc45:
The easiest solution would be to skip that test. I would only do that as a last resource, though. I wonder if this might be related to the tests that we are skipping for all E2E tests.

zhongruige:
Thanks @Luc45! The PR fixing it was closed as it [resolved itself](https://github.com/woocommerce/woocommerce/pull/39367#issuecomment-1654609598), so wonder if it's flakiness we're seeing here.

Luc45:
I see what's happening. We need to add the newly-added `filter-setter.php` mu-plugin that's part of the tests now. This is part of the test runner environment. I'll open a PR today.

zhongruige:
> I see what's happening. We need to add the newly-added `filter-setter.php` mu-plugin that's part of the tests now. This is part of the test runner environment. I'll open a PR today.

Oh interesting, good catch!

Luc45 (approved):

PR #72, opened 2023-08-23
Update self tests snapshots

zhongruige:


Luc45:
It's weird that they are all `pending`. Do they pass in WooCommerce?

MrJnrman:
At the time of the 8.1 release being cut, quite a few tests are being skipped which is why we are seeing them as pending. Unfortunately, there isn't much we can do about it.

MrJnrman (approved):
Adding my approval since the pending tests are expected since this is also what's happening in the WooCommerce core repo.

PR #75, opened 2023-09-04
Update to checkout@v4

Luc45:
As discussed in the Manager, this PR solves an issue with GitHub checkout@v3, flagged by the self tests.


Luc45:
Merging this as it has been approved on the Manager already

PR #78, opened 2023-09-19
Security Tests - Redirect rule tests

Luc45:
This PR is a counterpart to the QIT Manager's "Replace SemGrep redirect rule with PHPCS's" PR.


zhongruige (approved):
LGTM! Thanks for getting these tests updated.

PR #79, opened 2023-09-19
Run self-tests for Stable WooCommerce version as well

Luc45:
Our API tests were failing on Friday and Monday, and we didn't detect it because our self-tests runs against the `rc` version of WooCommerce.

As mentioned in the Manager PR, we assumed that all releases would be some kind of beta release before being made stable, but hotfixes are new releases that will go straight as a new stable, and a hotfix release threw our API tests off.

I've modified the self-tests to include tests for stable and RC versions, and to run twice a day so that we can act faster if needed.


zhongruige (approved):
Changes LGTM! Thanks @Luc45!

PR #82, opened 2023-09-27
Update self tests snapshots

zhongruige:

### DIFF
--- a/_tests/tests/__snapshots__/ApiTest__test_api_valid_features_stable_74__1.php
+++ b/_tests/tests/__snapshots__/ApiTest__test_api_valid_features_stable_74__1.php
@@ -1377,7 +1377,7 @@
         {
             "debug_log": [
                 {
-                    "count": "254",
+                    "count": "260",
                     "message": "PHP Notice: New Product Editor is enabled as expected. in \\/var\\/www\\/html\\/wp-content\\/plugins\\/woocommerce-product-feeds\\/woocommerce-product-feeds.php on line 12"
                 },
                 {
--- a/_tests/tests/__snapshots__/E2eTest__test_e2e_delete_products_rc_74__1.php
+++ b/_tests/tests/__snapshots__/E2eTest__test_e2e_delete_products_rc_74__1.php
@@ -27,7 +27,7 @@
             },
             "test_results_manager_url": "https:\\/\\/test-results-manager.com",
             "test_results_manager_expiration": 1234567890,
-            "test_summary": "Test Suites: 0 skipped, 4 failed, 46 passed, 50 total | Tests: 160 skipped, 7 failed, 35 passed, 202 total.",
+            "test_summary": "Test Suites: 0 skipped, 5 failed, 49 passed, 54 total | Tests: 159 skipped, 7 failed, 35 passed, 201 total.",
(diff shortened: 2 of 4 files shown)


Luc45:
Thanks for doing the sync, Greg. It seems the new Assembler Hub test is failing, though. :thinking: 

It seems we had a 404 on the post request to make this update: https://woo-compatibility-dashboard-tests.s3.amazonaws.com/compatibility-staging/e7350403-1f37-4d64-88a8-60cce4b04603-3f979c52-860a-42d0-b612-8555d4978430/index.html#/testresult/a5d229c8c406c848

I'd be curious to know if we need to do anything else to hook whatever code is needed to enable the endpoint on `wp-json/e2e-options/update`

Luc45:
Ah, it seems they added this to the e2e setup: https://github.com/woocommerce/woocommerce/blob/trunk/plugins/woocommerce/tests/e2e-pw/bin/test-helper-apis.php

This is tricky to track. I hope we don't need to patch things like this often. Ideally, the tests would be encapsulated or published in a standalone package for easier consumption. I'll be tweaking it to add that file to the sync.

zhongruige:
> This is tricky to track.

Wow nice catch on this, and agree, especially since there's [a couple of these now](https://github.com/woocommerce/woocommerce/tree/trunk/plugins/woocommerce/tests/e2e-pw/bin). This feels like something that could be turned into a plugin that is installed as part of the environment set up (and in general, thinking about it, this could be part of the utils packages that Solaris is working on).

Since these are formatted like plugins, I wonder if we can just do an install of any plugins we find within the `woocommerce/tree/trunk/plugins/woocommerce/tests/e2e-pw/bin` directory, to hopefully reduce some of the overhead with these, what do you think?

Luc45:
That same PR introduced some logic that invokes WP CLI from the context of the Monorepo, which breaks when we extract it outside of the Monorepo context, eg:

image

https://github.com/woocommerce/woocommerce/blob/trunk/plugins/woocommerce/tests/e2e-pw/utils/themes.js#L5

We can probably monkey patch this, but if this kind of change becomes common, it can truly jeopardize the QIT <-> WCCOM tests integration.

Luc45:
Ideally, those operations should be done using the UI in the browser.

zhongruige:
> We can probably monkey patch this, but if this kind of change becomes common, it can truly jeopardize the QIT <-> WCCOM tests integration.

I agree, we may just have to have a cut-off version at some point, otherwise we're going to get into a maintenance nightmare with these. I'll raise this, plus see if we can do anything to get early notification from them if there's any significant changes to how tests are being set up and ran so we can get ahead of issues like this, rather than during the update process.

Luc45:
I've monkey patched it successfully, the self-tests snapshots are currently being updated and I'll update them here soon.

I've dropped a comment tagging the specific code that caused some hiccups in this sync, asking to use the UI when possible in the future: https://github.com/woocommerce/woocommerce/pull/40235/files?w=1#r1343189857

Luc45 (approved):
Approved, all changes are as expected after monkey patching. Thanks for kicking off the sync, Greg!

zhongruige:
Thanks for getting in the fixes for this @Luc45!

PR #84, opened 2023-10-03
Normalize debug log in self tests

Luc45:
This PR improves our self-tests by:

- Removing some consistent PHP Debug Log errors that comes from WooCommerce Core or `wp-mail-logging` (used in E2E tests to mimick Solaris setup), as per [this PR](https://github.com/Automattic/compatibility-dashboard/pull/624).
- Allows some small variance in the count of PHP Debug Logs by rounding them to the closest number divisible by 10, so `234` errors become `230`, as to allow the tests to vary a little bit without making it a failure.
- Ignores PHP Debug Log errors that comes from `We couldn't connect to WordPress.org servers` as these can happen sometimes for reasons out of our control (like a network timeout between GitHub and WordPress.org, etc). We ignore it if it happens 3 times or less in a test (this is only for self tests snapshots).
- As an experiment, it keeps the PHP Debug Logs of E2E tests instead of removing them for the snapshot tests. We have removed them in the past because there was too much "chaos" going to on to snapshot test it, but with the normalization above, it might be possible to keep cover the PHP Debug Log of E2E tests in our self-tests. If these throw the self-tests off again, I'll disable it.

After merging this, let's see if the self-tests become less flaky. There's a small chance that I have to tweak this again. This flakiness only happens because we are very strict about our tests and we keep control of everything that happens in the PHP Debug Log - if we weren't so strict, this flakiness wouldn't happen. And it's not even flakiness - it's just because sometimes an Ajax request fires and sometimes not in an E2E test, depending on the speed on which the automated browser navigates away from the page, etc - so it's not flakiness, it's more that we are trying to keep it under the most control we can. Basically testing the boundaries here.

### DIFF
--- a/_tests/QITSelfTests.php
+++ b/_tests/QITSelfTests.php
@@ -243,17 +243,6 @@ function run_test_runs( array $test_runs ) {
 
 	echo "All tests finished. Processing results...\n";
 
-	$results = [];
-
-	foreach ( $processes as $p ) {
-		$results[] = [
-			'process'     => $p,
-			'has_results' => false,
-			'failed'      => null,
-			'succeeded'   => null,
-		];
-	}
-
 	$failed_tests = [];
 
 	/**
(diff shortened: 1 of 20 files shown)


zhongruige (approved):
Thanks for these changes @Luc45! Looks good to me, let's get this in and see if it helps with the flakiness we've been seeing.

PR #85, opened 2023-10-04
Update self tests snapshots

zhongruige:


zhongruige:
Closing-- the merge conflicts were because I didn't have the latest `trunk` locally and I think it'd be better to run this update with that.

PR #89, opened 2023-10-17
Update self tests snapshots

zhongruige:


Luc45 (approved):
I've regenerated the snapshots and deleted that binary file, not sure where that came from

zhongruige:
Same question I had too, was going to dig into it this morning but glad the re-generation fixed it. Super weird!

PR #92, opened 2023-11-08
Update self tests snapshots

zhongruige:


Luc45 (approved):
Some tests were failing due to 502 Bad Gateway when installing a plugin with WP CLI, I've implemented a retry logic and we are good to go.

Changes:
- `PHP Notice: Undefined property: wpdb::$use_mysqli` error is gone, this is expected as it's part of WordPress Core 6.4.1 hotfix
- Some line changes coming from WordPress Core upgrades
- Normalized paths `qit-runner` to `compatibility-dashboard` this is probably because the previous snapshot was generated in Prod or something like that, but I've checked and it's fine.

I'll be approving this and the one in the Manager.

zhongruige:
> Some tests were failing due to 502 Bad Gateway when installing a plugin with WP CLI, I've implemented a retry logic and we are good to go.

Nice! Thanks for digging into this and getting in that fix @Luc45!

PR #95, opened 2023-11-17
Update self tests snapshots

zhongruige:


Luc45 (approved):
Thanks, Greg. The notice `get_compatible_plugins_for_feature` is being generated on activation, but I'm not seeing it being called on our infra, so it seems to be external. I think we are good to go with the sync.

zhongruige:
> Thanks, Greg. The notice get_compatible_plugins_for_feature is being generated on activation, but I'm not seeing it being called on our infra, so it seems to be external. I think we are good to go with the sync.

Thank you for taking care of the merge and deploy!

PR #97, opened 2023-11-29
Added Theme self tests

MrJnrman:
This PR updates the self tests to include themes and also updates the snapshots to match updates results from the QIT manager

### DIFF
--- a/_tests/QITSelfTests.php
+++ b/_tests/QITSelfTests.php
@@ -22,7 +22,7 @@ class Context {
 
 	/*
 	 * To run tests in QIT, we need to assign the results to a plugin in the Marketplace.
-	 * We use the extension "woocommerce-product-feeds" and theme "storefront", because their owned by the test user in Staging.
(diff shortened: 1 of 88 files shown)


_tests/QITSelfTests.php
```diff
@@ -19,19 +19,19 @@ class Context {
 	public static $action;
 	public static $suite;
 	public static $test;
-	public static $sut_slug;
+
+	/*
+	 * To run tests in QIT, we need to assign the results to a plugin in the Marketplace.
+	 * We use the extension "woocommerce-product-feeds" and theme "storefront", because their owned by the test user in Staging.
```
zhongruige: Minor typo:

```suggestion
	 * We use the extension "woocommerce-product-feeds" and theme "storefront", because they're owned by the test user in Staging.
```

zhongruige (approved):
LGTM! Only one minor typo but otherwise the snapshot updates here look good 👍

PR #98, opened 2023-12-13
Sync Woo 8.4.0

Luc45:
- Changed ENABLE_HPOS to QIT_ENABLE_HPOS
- Updated the feature plugin we use to enable HPOS in a test site, as Woo 8.4 disallowed the way we were doing before
- Updated playwright to the latest version (same as Woo), as some tests were failing because it expected latest PW
- Ported over a feature plugin from upstream in API tests
- Skipped shipping tests in Woo >= 8.4.0 <8.4.1
- In the CLI, I've improved how we show the progress of parallel tests during self-tests, just so that we can have a better view if the self-tests are running as expected, as the linear blob of output was hard to parse

### DIFF
--- a/_tests/ParallelOutput.php
+++ b/_tests/ParallelOutput.php
@@ -1,0 +1,41 @@
+<?php
+
+use Symfony\Component\Process\Process;
+
+class ParallelOutput {
+	private $outputBuffer;
+	private $headers;
+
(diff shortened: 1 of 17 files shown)


zhongruige (approved):
LGTM! Thanks for updating and syncing the tests!

PR #99, opened 2023-12-15
Allow to use slug to list tests, color successes

Luc45:
This PR allows to use slugs instead of IDs when using `list-tests` command, it also add colors to status, and clean up the output a bit.

Example output on local environment:

`php qit-cli.php list-tests --extensions=woocommerce-product-bundles,woocommerce --test_types=api`

image

### DIFF
--- a/src/src/Commands/ListCommand.php
+++ b/src/src/Commands/ListCommand.php
@@ -46,7 +46,7 @@ View a list of test runs. You can also filter by extension, status, type, etc. E
 
 	qit list-tests --extensions=woocommerce,woocommerce-admin --test_status=success --test_types=security,e2e --page=1 --per_page=10
 HELP
-);
+			);
 	}
 
 	protected function execute( InputInterface $input, OutputInterface $output ): int {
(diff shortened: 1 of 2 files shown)


src/src/Commands/ListCommand.php
```diff
@@ -36,32 +36,43 @@ protected function configure() {
 		$test_types_list = implode( ', ', $this->environment->get_cache()->get_manager_sync_data( 'test_types' ) );
 		$this
 			->setDescription( 'List test runs.' )
+			->addOption( 'extensions', 'e', InputOption::VALUE_OPTIONAL, '(Optional) Retrieve results for these extensions (Accepts slugs or IDs, comma-separated).' )
```
zhongruige, line 39: Minor suggested tweak:

```suggestion
			->addOption( 'extensions', 'e', InputOption::VALUE_OPTIONAL, '(Optional) Retrieve results for these extensions (Accepts either a comma separated list of slugs or IDs).' )
```

Only because as it's currently worded someone might think they could do something like this:

```
php qit-cli.php list-tests --extensions=woocommerce-product-bundles,woocommerce,4652610 --test_types=api
```

Which will only run the list against the ID.

Other than that, this tested great and I think it's good to go!

Screenshot 2023-12-15 at 8 48 55 AM

Luc45: It actually accepts both IDs and slugs
image

zhongruige: Ah, thanks, I switched to prod and it worked, so much be something odd in staging that was causing that behavior.

zhongruige (approved):
LGTM! :shipit:

PR #101, opened 2023-12-20
Add jetpack scan self test

MrJnrman:
This PR removes the old malware scanner self test and replaces it with the jetpack malware scan


zhongruige (approved):
LGTM! Thanks for getting in this self-test update @MrJnrman!

Luc45 (approved):

PR #103, opened 2023-12-21
Use zip from docker

Luc45:
All self-tests are failing because GitHub changed something on their `zip` binary. This PR changes the implementation to use Zip from Docker instead.

Example of passing job: https://github.com/woocommerce/qit-cli/actions/runs/7288764268/job/19861964831


zhongruige:
Thanks for jumping on this @Luc45!

PR #105, opened 2024-01-09
Sync 8.5.0

Luc45:
- This PR updates the self-test snapshots for Woo 8.5.0
- Minor tweak to `make build`
- Increase default test timeout from 30m to 60m


zhongruige (approved):
Thanks for making these updates and increasing the timeout here @Luc45! LGTM :shipit:

PR #109, opened 2024-01-11
Unify run_id  header in outputs.

tomalec:
Unify header used for run id in `--wait` and regular mode.
So it's easier to interpret the results.

### Before
Currently, a call with `--wait` returns, but calling it without calls is the same thing as "Test ID".
```sh
$ qit run:activation google-listings-and-ads

Running test...
Test started on the QIT Servers!
Test ID    4547685                                                                                                                              
…

$ qit run:activation google-listings-and-ads --wait

Running test...
Test run completed.
Run Id               4547687                                                                                                                              
…

$ qit run:activation google-listings-and-ads --wait --json
{"run_id":4547688,…
```

### After
Both return
```
Run Id
```

## Alternatives to consider.

`qit get` call it even more verbosely "test run id", so maybe we should stick to that.
```sh
qit get --help
Description:
  Get a single test run.

Usage:
  get [options] [--] <test_run_id>

Arguments:
  test_run_id   
```

## Context

I'm trying to add a bit more GHactions-based automation to interpret the output and pass it to the following actions and workflows.
Having only one format to parse would make it a bit easier.

Also, it would be nice if I could call it w/o `--wait` but with `--json` and get the test run ID value.


Luc45:
howdy @tomalec, thanks for the feedback.

I agree we could normalize `Test ID`, `Run ID` into `Test Run ID`, both in the text output and in arguments where applicable.

However, any programmatic integration you do based on the output should ideally be done with the `--json` flag. The non-JSON output is considered just text oriented for the user and is more likely to be changed, whereas the JSON output is more likely to remain unchanged, which makes any programmatic integration you build on top of it more solid.

You can use `jq` to parse the JSON output that you receive back, eg:

1. Running a security test and getting the result ID: `qit run:security some-extension --json --wait | jq '.run_id'` (Returns `123456`)
2. Getting the status:  `qit get 123456 --json | jq '.status'` (Returns `"failed"`)

tomalec:
> However, any programmatic integration you do based on the output should ideally be done with the `--json` flag


Thanks for the hint. I'd definitely do that; the problem is `quit run:security some-extension --json` (without `--wait` returns, no JSON whatsoever.

I think that's another area for improvement, but it's a bit more complicated for me to do that this PR ;)

Luc45:
Got it, I'll make it return the async response in that case. PR should be up by EOD.

Luc45:
@tomalec Can you please pull this branch and see if that solves your use-case?

You can try the development build in this branch with either option:

- `cd this-repo` && `./qit run:activation automatewoo --json`
- or `cd this-repo/src` && `php qit-cli.php run:activation automatewoo --json`

tomalec:
Awesome, thanks! Yes, it does solve my use case

Luc45:
Now `test_run_id` is used more consistently, and it's possible to get the JSON response without waiting for the test to finish:

#### Starting a test run:
image

#### Getting a test run:
image

#### Getting a test run JSON (Property `test_run_id` was added, I kept `run_id` for backwards compatibility, will be removed later):
image

#### Running a test with `--json` flag without waiting for the results:
image

#### Running a test with `--json` flag and `--wait`:

image

zhongruige (approved):
Changes look good to me and I was able to kick off test runs from the CLI as expected. :shipit:

tomalec:
Thank you for taking over the PR and pushing it so much further!

PR #113, opened 2024-01-17
Ignore unescaped exceptions

Luc45:
We recently upgraded from WPCS 2 to WPCS 3, which adds a new assertion that is prone to false-positives, as per convo on https://a8c.slack.com/archives/C03R7A7D1DW/p1705492174805429

This PR ignores the new unescaped exception rule in WPCS 3.


Luc45:
Merging as approved upstream

PR #114, opened 2024-01-18
Debug failed self-tests

Luc45:
This PR aims to improve self-tests, which are failing randomly.

- It includes non-JSON text to the parallel process output (Any non-JSON output generated by the process running the test was previously discarded)
- It saves the JSON output (which is the result JSON) in a file and reads from the file instead
- It preserves output generated by the orchestrating process
- It uses a unique filename for result JSONs, which in rare cases could cause a race-condition
- It adds a `--debug` flag that shows all remote requests done (not enabled in CI for now)

This makes the self-tests more robust, and helps to debug it if it fails again.

### Testing instructions
- Pull this branch
- Point your CLI to Prod
- Go to `_tests` directory and run `php QITSelfTests.php run` (Optionally add `--debug`)
- Wait for all tests to complete
- Assert all tests completes successfully


Luc45:
I'll be going ahead and merging this to reduce PR review churn, since it's just related to self-tests, and the snapshots shows that there are no changes to results.

PR #116, opened 2024-01-24
Update QIT CLI to the new domain

Luc45:
This PR:

- Updates the QIT CLI to use `qit.woo.com`
- Adds [CURLOPT_POSTREDIR](https://curl.se/libcurl/c/CURLOPT_POSTREDIR.html) so that curl maintains the original HTTP method when encountering redirects (essentially keeping the POST HTTP method after encountering a 301~ redirect - without it, the requests are received on the redirected server as GET)
- Adds verbose curl output if verbose mode is enabled
- Updates the old domain to the new domain

Testing instructions:

- Pull this branch
- Interact with QIT Manager (which is still on the old domain)
- Enable verbose mode (`-vvv`)
- Assert that it is trying to communicate with qit.woo.com but being redirected to the old domain, and everything still works

This can be released even before we switch to `qit.woo.com`. Teams that upgrade early will have zero downtime on their CI.


zhongruige:
Thanks for this at @Luc45! Just want to confirm, I'm not seeing the redirect. I tried a build and using `php qit-cli.php run:activation`, but the results didn't show the `qit.woo.com` domain in the output, just the original domain, so want to confirm if this is still expected or if I should see the new domain there. Thanks!

Luc45:
This is expected, it should show the actual QIT manager domain (the old one), even though it's communicating with the new one, still unreleased.

zhongruige:
> This is expected, it should show the actual QIT manager domain (the old one), even though it's communicating with the new one, still unreleased.

Awesome, thanks for confirming!

zhongruige (approved):
Changes LGTM! Thanks for getting this in!

PR #117, opened 2024-01-25
Add self-test for PHPCompatibility to not flag attribute usage

Luc45:
As discussed in Slack, attributes are usually backwards compatible with PHP 7, for the exception or multi-line attributes.

PHPCompatibility currently cannot differentiate between them and flags all usage of Attributes as invalid if you plan to support PHP 7.

We disabled the Attribute rule in the PHPCompatibility test as it's currently causing more false-positives than bringing benefits, as multi-line attribute usage is very rare in WP plugins, but `ReturnTypeWillChange` is very common, for instance.

We are keeping an eye upstream for when PHPCompatibility have a way to [differentiate](https://github.com/PHPCompatibility/PHPCompatibility/issues/1480) both, and will re-enable this rule.

- We can see that once I added an attribute to the self-test, the attribute was first [flagged](https://github.com/woocommerce/qit-cli/commit/ca7ba659e61274de212389cdaf3dd8d36e03bc65)
- Then we ignored the Attribute rule in QIT
- Then the attribute is [no longer flagged ](https://github.com/woocommerce/qit-cli/commit/4149c111fa65eea3761a26aec2f19e6f77410cc9)

### DIFF
--- a/_tests/phpcompatibility/sut_default_php_version/woocommerce-product-feeds/woocommerce-product-feeds.php
+++ b/_tests/phpcompatibility/sut_default_php_version/woocommerce-product-feeds/woocommerce-product-feeds.php
@@ -29,6 +29,22 @@ class Bar {
 	readonly string $foo;
 }
 
+/*
+ * Attributes require PHP 8, but we disabled this check for now because
+ * it is generally backwards compatible, it's only invalid syntax in PHP 7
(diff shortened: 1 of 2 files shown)


zhongruige (approved):
Thanks for getting this in @Luc45!

Issue #108, opened 2024-01-11
Rename "application_password" to "qit_token"


Luc45:
The current setup in QIT already incorporates a unique and secure approach to authentication, but the terminology used, `application_password`, can lead to confusion as indicated in this GitHub issue: https://github.com/woocommerce/qit-cli/issues/106

**Clarifying Purpose and Usage:** The term `application_password` is a generic one in the WordPress context and can be used as a password applicable for various applications or endpoints within WordPress. Renaming it to `qit_token` immediately clarifies that this token is specifically for use with QIT. It removes ambiguity, making it clear to users that this token is not a standard WordPress application password but something specific to QIT.

### What's different in a `qit_token`?

- **Exclusive Authentication for QIT:** The `qit_token` is a unique application password specifically designed for authenticating requests against QIT endpoints.
- **Limited Scope of Use:** A `qit_token` is intentionally restricted and cannot be used to authenticate requests against non-QIT endpoints.
- **Restriction on Endpoint Access:** Unlike regular application passwords in WordPress, QIT endpoints are configured to only accept the `qit_token` for authentication.
- **Enhanced Security Measures:** The use of `qit_token` significantly increases security. In the event of a token leak, its impact is confined to QIT endpoints only. This restriction is crucial as it prevents the compromised token from being used to perform potentially harmful actions outside of QIT, such as modifying products or accessing sensitive data.

In summary, while the backend functionality and security measures are already in place and effective, renaming `application_password` to `qit_token` in the QIT CLI commands and documentation will improve clarity and user experience.

Luc45:
Solved in #118

PR #118, opened 2024-01-26
Rename "Application Password" to "QIT Token" and other improvements

Luc45:
This PR aims to reduce authentication confusion by renaming `application_password` to `qit_token`.

Solves https://github.com/woocommerce/qit-cli/issues/108

It also does some other minor improvements, such as moving examples on how to run tests to AFTER authentication is complete.

It keeps the `--application_password` option for backwards compatibility for the time being.

Testing Instructions will be added in Docs repo.


zhongruige (approved):
Tested great for me, thanks for these updates @Luc45!

PR #123, opened 2024-02-05
Allow retry on get request

Luc45:
Self-tests failed today, upon investigating it seems the request to get the result failed due to 429 (too many requests). This PR allow this request to be retried, which should solve it.


zhongruige (approved):
Thanks for catching this and adding in this retry logic @Luc45 !

PR #125, opened 2024-02-09
Rename "api" to "woo-api" and "e2e" to "woo-e2e"

Luc45:
On current versions of QIT CLI, they will see both the renamed and the alias:

image

Once they update to the latest version, they will see just the renamed:

image

The aliases are still there, but they are hidden. If they run it (as in a CI script that still uses  `e2e` as the test-type), it still works:

image

If the user starts a test using `e2e` as the test type, instead of `woo-e2e`, internally we route it to the renamed test type and log it to keep track of usage:

image

In the future, when `e2e` actually refers to custom tests, we will:

1. Check if the SUT has custom tests
1.1 If yes, we just run it
1.2 If no, this can be either because the dev hasn't added custom tests yet, or because it's a old CI script still pointing to this test type
2. If `getenv('CI')` is true, we just run the `woo-e2e` test type and add a warning to the test
3. If false, we bail with a error instead, and point the dev towards documentation on how to write custom tests

## Testing Instructions

Testing instructions will be added to the Manager PR.


src/src/Commands/CreateRunCommands.php
```diff
@@ -306,6 +306,24 @@ public function execute( InputInterface $input, OutputInterface $output ) {
 
 		$command->add_option_to_send( 'zip' );
 
+		if ( $test_type === 'api' ) {
```
zhongruige, line 309: I like this so we can always come back and add custom API tests in the future.

src/src/Commands/CreateRunCommands.php
```diff
@@ -306,6 +306,24 @@ public function execute( InputInterface $input, OutputInterface $output ) {
...
+			$command->setHidden( true );
+		}
+
+		// This will be hidden while custom test type is in development, but kept as an alias to "woo-e2e".
+		if ( $test_type === 'e2e' ) {
+			try {
+				$hide_e2e = $this->environment->get_cache()->get_manager_sync_data( 'hide_e2e' );
```
zhongruige, line 316: I like how this configuration is set.

zhongruige (approved):
Nice work on this @Luc45! This refactor helps to provide us separation of concerns for the different test types and sets us up for the next stages in QIT. Self tests all passed:

Screenshot 2024-02-09 at 3 32 59 PM

PR #126, opened 2024-02-09
Use dev version of symfony/process

Luc45:
The [Symfony Process PR](https://github.com/symfony/symfony/pull/53821) was merged upstream, by looking at their release cycles it seems it should be released as a stable version by the end of the month.

We can use it immediately with `minimum-stability: dev`, which this PR does.


zhongruige (approved):
Tested great for me, thanks for this update @Luc45!

Screenshot 2024-02-09 at 9 17 15 AM

PR #131, opened 2024-02-27
Add self-tests for Activation Tests loopback

Luc45:
PR comments upstream.


zhongruige (approved):
LGTM! Thank you for adding in these self tests @Luc45!

PR #133, opened 2024-02-28
Custom E2E Tests 1 - Renaming `Environment` to `Backend`

Luc45:
This PR prepares the codebase for the Custom E2E tests and local test environment features.

- Make `Cache` a singleton in the DI container, so that whenever an instance is requested, it will return the same, known instance throughout the entire request
- This made it possible to access `Cache` directly without getting it from `Environment`, which reduced the usage of `Environment` on the codebase to just the `add`, `remove` and `switch` commands, whereas previously it was everywhere we needed to read from cache.
- Renamed `Environment` to `Backend`. This class is basically the connection to the Manager Backend. It was called `Environment` because you could connect to a Prod backend, staging, local, etc. I needed the `Environment` name for the local test environment, so I renamed this to `Backend` to reduce confusion and reduce ambiguity in the codebase.

I'll be marking it as draft until we can merge all Custom E2E Tests PR.


zhongruige (approved):
Thanks for making this change! Verified that I could still run tests with the renaming and impacting any of my current flows.

PR #134, opened 2024-02-29
Fix QIT config path in Windows

Luc45:
On windows specifically there was a typo that would save config files in `C:\Users\AppData\foo\woo-qit-cli.env-default` instead of `C:\Users\AppData\foo\woo-qit-cli\.env-default` (Missing the final `\` there)

This PR addresses it, and makes the Config function more uniform.

It also ignores the `/dev` dir to avoid conflicts while working in the Local Test Environment branch and switching back-and-forth.

### Testing instructions
- Pull this branch
- Interact with QIT, nothing should change
- If you have a Windows VM around, try it in there, it should work as well


zhongruige:
Tested this going through the default (no changes to auth) flow on macOS and I was able to create new test runs and, after renaming the existing directory, it created the new directory as expected and I was able to create new test runs, too.

zhongruige (approved):
Thanks for digging into this and getting this fix!

PR #135, opened 2024-03-01
Update snapshots 8.7.0-beta.2

Luc45:
PR comments upstream.

I skipped the PHP debug log on `delete_products` tests of API tests, as they are becoming hard to predict with such accuracy.

In this test, we add a mu-plugin that deletes a product after being created, to assert that the test fails when it should. Depending on what happens on the request (ajax, cron, etc), this might generate different PHP debug logs and we don't need to have this level of control for this test type, we just need to assert that it fails as it should, not that the debug logs should look like exactly like a previously saved snapshot.


zhongruige (approved):
Thanks for getting in these test updates @Luc45!

PR #137, opened 2024-03-05
Retry after 429

Luc45:
If we encountered a 429 before, it would be handled as a "generic failure" and depending on the request, would be retried.

Now, we handle 429 in a special way, we first look if the server sends a `Retry-After` header and use that, and if not, we implement an exponential back-off that will wait increasingly longer times between retries.

To test this PR, I've removed all "wait between tests" logic from the self-tests and I can assert that they all can be dispatched simultaneously and succeed. (E2E tests were still running when I took the screenshot)

image

### Testing Instructions
- Pull this branch
- Run some tests and assert they succeed
- Run the self-tests and assert that it tries to dispatch all of them in parallel, and succeeds


zhongruige (approved):
Changes look good to me! Thanks for updating this retry logic @Luc45!

PR #138, opened 2024-03-05
Windows SSL

Luc45:
PHP on Windows comes by default without a Certificate Authority file, which prevents it from making any network request to URLs with `https://` (or maybe, to only a few, as it seems it succeeds at https://google.com)

This PR adds a fallback to use a CA file on Windows if a request fails.

I've opened a feature request on PHP asking for this to be fixed: https://github.com/php/php-src/issues/13601

And also did some [isolated](https://github.com/Luc45/ci-playground/actions/runs/8159698661/job/22304587825) tests on a dedicated repo.


Luc45:
We can see it working in this test: https://github.com/woocommerce/qit-cli/actions/runs/8162261104 (Updated)

What we do here is:

- Using `windows` in CI, for each PHP version (7.2 to 8.3)
- We try to sync without any Certificate Authority and see it fails
- Then we allow it to download the CA file (on a Windows desktop we would ask the user)
- We repeat the request and assert it works
- We repeat it again to test a cache hit (CA file already downloaded) and assert it works

This logic only triggers when running on Windows. (I've used the _"QIT Environment Test - Linux_" workflow just to do my tests on this branch)

.github/workflows/qit-windows.yml
```diff
@@ -0,0 +1,78 @@
+name: QIT Windows
```
zhongruige, line 1: Nice, thanks for adding in this test workflow!

Luc45:
Manual test on native Windows (Second `yes` is cached)

image

zhongruige (approved):
LGTM! Thanks for getting this in @Luc45! I ran some tests locally and things continued to work as expected on macOS.

PR #139, opened 2024-03-07
Use Composer CA Bundle instead of custom implementation

Luc45:
Turns out that Composer also faced the SSL issue we faced on Windows and released their solution as a small package.

This PR removes our custom implementation of the fix, which works, to use Composer's instead, which is more robust, has been battle-tested trillions of times, and has the advantage that it works without asking the user if he wants to download the CA file. I've debated bundling the CA file instead of downloading it, but I didn't want to keep it updated, that's why I decided to ask the user to download it at runtime, but since this is handled by the package for us, then it's better.

Relevant links:

- https://github.com/composer/ca-bundle
- https://packagist.org/packages/composer/ca-bundle


Luc45:
We can see on [QIT Windows workflows](https://github.com/woocommerce/qit-cli/actions/runs/8181564116) that it used the bundled cert

image

I've tested on a Windows laptop and it also worked.

zhongruige (approved):
Thanks for staying so on top of this @Luc45 and getting these changes in. This is a good approach to this problem here :shipit:

PR #146, opened 2024-03-14
Add self-tests for `WordPress.DB` in Security Tests

Luc45:
These are the self-tests for the upstream PR that adds `WordPress.DB` sniffs to Security Tests.


zhongruige (approved):
LGTM! Thanks @Luc45!

PR #150, opened 2024-03-20
Update malware tests

MrJnrman:
Update malware self tests to omit suspicious links marked as `testing`


zhongruige (approved):
Thanks for these updates!

PR #152, opened 2024-03-21
Add empty extension list exception

zhongruige:
This PR adds a small change with an empty extension list exception, and returning this error in the catch block if we hit it.


Luc45 (approved):
Thanks! :+1:

PR #158, opened 2024-04-04
Update WC 8.8 rc test snapshots

zhongruige:


Luc45 (approved):
Thanks Greg! I've pushed a commit to skip 2 flaky tests (that was there since last sync with Stable, unrelated to this PR). I think we should also bring up this deprecation warning upstream. Other than that, merge away! :shipit:

zhongruige:
> I've pushed a commit to skip 2 flaky tests

Thanks!

> this deprecation warning upstream

Good point, I can let the teams know we saw this while updating our snapshots.

PR #163, opened 2024-05-03
Add `woo:validate-zip <path>` command

rcstr:
Part of https://github.com/Automattic/woocommerce.com/issues/20009

### Description
PR adds a basic zip file validation command. Validation is based on:

https://github.com/Automattic/woocommerce.com/pull/20050/files#diff-d1f39d6ef6800717504f5c7dffecf7a012b52659250ba19e7b84f5dbba532c83R120-R126

### Testing instructions
1. Checkout branch
2. Use https://github.com/Automattic/woocommerce.com/pull/20050/files?show-viewed-files=true&file-filters%5B%5D=#diff-839d4fbb14d0686c8ed4cc751e54a9d9a5b586bfabc494cfcd68f0ff53b27ed4 and https://github.com/Automattic/woocommerce.com/pull/20050/files?show-viewed-files=true&file-filters%5B%5D=#diff-36f515fa8eb2124c6844a1ab812d106ab7c677b1973b19e03c721704a518fa39 folders to generate invalid zip files
3. Run `make build`
4. To run the command, use `woo:validate-zip <path>`. Where `<path>` is the path for the zip file, eg `woo:validate-zip /Users/rommel/Projects/woocommerce.com/tests/integration/data/plugins/invalid-file-inside.zip`

### Screenshots
Screenshot 2024-05-03 at 04 07 09


Luc45 (changes requested):
howdy @rcstr , thanks for putting this together!

I'll do a comprehensive code review pass with my impressions, some of them are important, some of them can be considered minor nitpicks, and I'll group all these feedback in an objective, straightforward list:

- First, I would put this command in a new, dedicated namespace called `woo:validate-zip`, where this namespace will be for WooCommerce.com Marketplace-specific operations. This way, we keep the `run` namespace for test types, which are more often than not, independent from the Marketplace. Our vision for the long-term is to open QIT to other extensions as well, such as those on WordPress.org, so keeping Marketplace-specific code containerized helps us achieve that goal.
- Similarly, I'd move the `Zip` class from the parent namespace to a `Woo` namespace.
- I'd rename the `Zip` class to `ZipWooValidator` or similar, to better denote the purpose of the class just by glancing at it.
- Can you please clarify why static is needed on the `Zip` class?
- If we remove statics, we can add the `Zip` class to the constructor of the `RunValidateZipCommand`, leveraging the DI container.
- Just something to keep in mind: This check on the `Zip` class: `$opened = $zip->open( $filepath, ZipArchive::CHECKCONS );` might fail zip files that are created using MacOS Archive Utility. I'm not sure how to deal with this, I guess it depends whether this is a requirement upstream (on the Marketplace) as well. Ultimately, zips created by the MacOS Archive Utility can be installed in WordPress with no problems, as WordPress Core will fallback to pclzip when that ZipArchive checkcons check fails, so should we block it based on that? Debatable. But maybe we could inform the user with a meaningful message?

rcstr:
@Luc45 - thanks for the review. I have handled your points:

- Renamed command to `woo:validate-zip` c9c5b43
- Move `ZIP` to `QIT_CLI\Woo\ZipValidator` class and make it a service 1c68d32
- Added message for _Archive Utility_ possibly created files 6ba3ed8

src/src/Woo/ZipValidator.php
```diff
@@ -0,0 +1,125 @@
...
+	 *
+	 * @return \ZipArchive
+	 * @throws \RuntimeException If the file is not a valid zip file.
+	 */
+	private function open_file( string $filepath ) {
+		$zip    = new ZipArchive();
+		$opened = $zip->open( $filepath, ZipArchive::CHECKCONS );
+
+		if ( $opened === 21 ) {
+			throw new \RuntimeException( 'Zip file is inconsistent. Zip files generated by the Archive Utility from macOS may be the cause.' );
```
Luc45: A nitpick, but here we are failing any zips generated by MacOS Archive Utility.

What about instead, we do this?

If it fails the consistency check in a way that matches those zips generated by MacOS Archive Utility, we proceed with the script and show a warning to the user.

```php
	private function open_file( string $filepath ) {
		if ( $opened === 21 ) {
			throw new \UnexpectedValueException( 'Zip file is inconsistent. Zip files generated by the Archive Utility from macOS may be the cause.' );
		}
	}

	public function validate_zip( string $filepath ) {
		try {
			$zip = $this->open_file( $filepath );
		} catch( \UnexpectedValueException $e ) {
			// Let the user know if it fails consistency check because it was generated on Mac, but proceed.
			// Any other exception (such as RuntimeException) will not be caught and will interrupt the script.
			\QIT_CLI\App::make( \QIT_CLI\IO\Output::class )->writeln( sprintf( '<comment>%s</comment>', $e->getMessage() ) );
		}
	}
```

rcstr: Implemented in 6ba3ed8...060fa17

Screenshot 2024-05-09 at 15 10 49

docs/dev.md
```diff
@@ -0,0 +1,15 @@
+### `detected dubious ownership in repository` error during `make build
```
zhongruige: Thanks for adding in this doc update! Only a minor adjustment:

```suggestion
### `detected dubious ownership in repository` error during `make build`
```

src/src/Commands/WooValidateZipCommand.php
```diff
@@ -0,0 +1,48 @@
...
+
+			$this->zip_validator->validate_zip( $zip_file );
+
+			$output->writeln( '<info>Zip file content is valid.</info>' );
+		} catch ( \UnexpectedValueException $e ) {
+			$output->writeln( sprintf( '<comment>%s</comment>', $e->getMessage() ) );
+
+			return Command::FAILURE;
+		} catch ( \Exception $e ) {
+			$output->writeln( sprintf( '<error>An error occurred while validating the ZIP file. Error: %s</error>', $e->getMessage() ) );
```
zhongruige: Minor wording change suggestion:

```suggestion
			$output->writeln( sprintf( '<error>An error occurred while validating the provided file. Error: %s</error>', $e->getMessage() ) );
```

src/src/Woo/ZipValidator.php
```diff
@@ -0,0 +1,125 @@
...
+				if ( ! $info ) {
+					throw new \RuntimeException( 'Cannot read zip index.', 400 );
+				}
+
+				if ( str_contains( $info['name'], '/vendor' ) || str_contains( $info['name'], '/node_modules' ) ) {
+					continue;
+				}
+
+				if ( $this->is_file_invalid( $info['name'] ) ) {
+					throw new \RuntimeException( sprintf( 'Invalid (%s) file/folder inside zip file', $info['name'] ), 400 );
```
zhongruige: Minor wording change:

```suggestion
					throw new \RuntimeException( sprintf( 'Invalid (%s) file/folder inside the provided ZIP file', $info['name'] ), 400 );
```

src/src/Woo/ZipValidator.php
```diff
@@ -0,0 +1,125 @@
...
+
+		$left  = 0;
+		$right = $zip->numFiles - 1; // phpcs:ignore WordPress.NamingConventions.ValidVariableName.UsedPropertyNotSnakeCase
+
+		while ( $left <= $right ) {
+			foreach ( [ $left, $right ] as $i ) {
+				$info = $zip->statIndex( $i );
+
+				if ( ! $info ) {
+					throw new \RuntimeException( 'Cannot read zip index.', 400 );
```
zhongruige: We use both `zip` and `ZIP` interchangeably, so we may want to settle on one. I think `ZIP` is the general standard, so maybe we can settle on that?

zhongruige (commented):
So far this has been testing great for me with both valid and invalid zip files, nice work. I added a few comments around some suggested wording changes, let me know your thoughts!

src/src/Woo/ZipValidator.php
rcstr: updated `zip/Zip` to `ZIP` in c181a53

zhongruige (approved):
Ran through the changes again after the merge conflicts were resolved and everything tested great again for me:

Screenshot 2024-05-13 at 4 06 52 PM

Nice work on this @rcstr! After @Luc45 reviews the changes he had originally requested are resolved, we should be all set.

Once this is done, I believe we can get this [added to our documentation](https://qit.woo.com/docs/category/managed-tests), I think under "Managed Tests" might make the most sense, and we can mention the files and folder structures this checks for to provide more context for the user, then in a follow up task, add a link to the docs in the results. Let me know your thoughts!

PR #166, opened 2024-05-10
Tweaks to release process

Luc45:
This PR proposes some tweaks to the release process.

Instead of inferring the version from the tag, we set the version and then create the tag.

This reduces some friction when building on Mac, and makes the release process more straightforward and the release.md easier to follow. (We removed `"git-tag": "QIT_CLI_VERSION",` from box.dist.json)


zhongruige (approved):
Nice! Thanks for getting this in @Luc45, tested great for me:

Screenshot 2024-05-10 at 10 34 07 AM

cc'ing in @rcstr as with this change we can [remove the README update](https://github.com/woocommerce/qit-cli/pull/163/files#diff-3a6cab76ecce19612a704fe0ef89dc83f1ed2df44f7bfa28d47d2529a7949fafR1) for this since it won't be an issue anymore going forward.

PR #167, opened 2024-05-14
Upload test artifacts to debug self-tests flakiness

Luc45:
There is a constant flakiness on the self-tests for the custom tests, this PR uploads the test artifacts to the workflow so that we can download and view the report locally.


zhongruige (approved):
Changes LGTM, hopefully we can figure out what the root cause of the flakiness is

PR #177, opened 2024-06-17
Fix issue with "--plugins=woocommerce" and "--woo" parameter

Luc45:
As per Slack https://a8c.slack.com/archives/C03R7A7D1DW/p1718530807725489

Fixes an issue on running Custom E2E tests when you used both the `--woo` parameter and `--plugin=woocommerce`. If you used `--plugin woocommerce` it would work.

This PR fixes this behavior, and also prevents the user from using both `--woo` and `--plugin=woocommerce`, and asks to pick one or the other.

### Testing Instructions
- Check out this branch
- Try to run a test with both `--woo <version>` and `--plugin woocommerce` and see that it prevents you from doing that
- Check that the self-tests for the custom tests are passing:
  - _PS: You might need to spin up your Manager or deploy trunk to Staging depending on what environment you have in your `.env` file._
  - `cd _tests/custom-tests`
  - `./vendor/bin/paratest`


src/src/Commands/CustomTests/RunE2ECommand.php
```diff
@@ -146,6 +146,17 @@ protected function execute( InputInterface $input, OutputInterface $output ): in
 		$source              = $input->getOption( 'source' ) ?? $woo_extension;
 		$sut_action          = $input->getOption( 'sut_action' );
 
+		// Prevent usage of "--woo" and "--plugin woocommerce" together.
+		if ( ! empty( $woocommerce_version ) && ! empty( $input->getOption( 'plugin' ) ) ) {
+			foreach ( $input->getOption( 'plugin' ) as $p ) {
+				if ( $p === 'woocommerce' ) {
+					$output->writeln( '<error>Cannot use "--woo" and "--plugin woocommerce" at the same time. Use one or another.</error>' );
```
zhongruige: Minor suggestion on different messaging for the error:

```suggestion
					$output->writeln( '<error>Both "--woo" and "--plugin woocommerce" cannot be used at the same time. Please use one or another when running the test.</error>' );
```

zhongruige (approved):
Thanks for getting in this quick fix @Luc45! Tested great, just a minor comment on a wording change for the error message, otherwise the changes looks good to me.

PR #178, opened 2024-06-17
Use Input instead of argv, activate Woo instead of bootstrapping when using "--woo" parameter

Luc45:
This PR is a continuation of [this PR](https://github.com/woocommerce/qit-cli/pull/177), with an additional fix that I forgot to include in that PR.

Previously, when the `RunE2ECommand` triggered the creation of a test environment via `UpEnvironmentCommand`, the latter would read command-line parameters directly from `$argv`, which contains user-defined input. The issue arose because `UpEnvironmentCommand` was inadvertently using the `$argv` parameters intended for `RunE2ECommand`, since both commands were executed within the same process.

To resolve this, this PR modifies `UpEnvironmentCommand` to accept options through a dedicated variable. If this variable is set, `UpEnvironmentCommand` will use these values; otherwise, it will fall back to `$argv`. This ensures correct parameter usage, particularly when environments are set up independently of the `RunE2ECommand`.

### Testing Instructions
- Self-tests for the custom tests should pass
- Scaffold a sample test somewhere: `qit scaffold:e2e /tmp/foo`
- Run that sample test with `--woo=8.6.2` for instance: `qit run:e2e automatewoo /tmp/foo --php_version=7.4 --wp=6.5.2 --woo=8.6.2`
- See that the test runs normally (the issue was with `--woo` parameter overriding the plugins list)

### Screenshots

**Test passing as instructed**

image

**Self-tests passing**

1 snapshot regenerated, which is the one in this PR.

image


zhongruige (approved):
Changes tested great on this, thanks for getting this in @Luc45!

PR #179, opened 2024-06-18
Update snapshots for Security Tests

Luc45:
Takes into account Minimum WP Version to stop flagging usage of `%i` placeholder on WP < 6.2, since Woo now requires 6.4+.


zhongruige (approved):
LGTM!

PR #185, opened 2024-07-17
Wait 10 seconds between each self-test test

Luc45:
This PR extracts the `ProcessManager` composer dependency (which was a single file) into it's own class so that we can modify it.

It adds a configurable wait between each process start. We use this to wait 10 seconds between each test during self-tests to make it more robust.

This makes the test wait a little bit between each self-test instead of starting them all at once.


zhongruige (approved):
Thanks for getting this change in @Luc45! The configurable wait is a really nice addition to this.

PR #190, opened 2024-07-31
Temporary test environment tweaks

Luc45:
As per convo on Slack:

- Respect mount flag if user specified one
- Expand relative volume to be relative from the cwd
- Make sure mu-plugins dir exists and has expected permissions


MrJnrman (approved):
Thanks for the quick turnaround here Lucas! All changes work as expected without any impact to the original functionality.

PR #195, opened 2024-08-15
Support qm logs

MrJnrman:
WIP

Added support for collection Query Monitor logs

### Changes in this PR
- If the QIT_TEST_RUN_ID var is present, then send it along with the test run started notification
- Added a PrepareQMLog class that prepares the QM logs.
- If there is logs folder in wp-content/plugins then we send a summary of that data with the test run ended notification.

**Note**:
If no QM log exists, then the original prepared debug log is sent, if the QM log does exists then its sent along with the original prepated debug log.

### Structure of debug log returned to the Manager
**If QM Logs Exists**
```
{
        "debug_log": "<original_debug_log_json>",
	"qm_logs": {
		"non_fatal": "<qm_log>",
		"fatal": "<debug_log_fatals>",
	}
}
```

**If QM Logs Not Exists**
```
{"<original_debug_log_json>"}
```


zhongruige (approved):
Nice work on this @MrJnrman!

PR #196, opened 2024-08-22
Fix --source option

Luc45:
This PR fixes the `--source` option on `run:e2e` as discussed in Slack.

### DIFF
--- a/src/src/Commands/CustomTests/RunE2ECommand.php
+++ b/src/src/Commands/CustomTests/RunE2ECommand.php
@@ -193,8 +193,8 @@ class RunE2ECommand extends DynamicCommand {
 			// Validate WooExtension.
 			try {
 				if ( is_numeric( $woo_extension ) ) {
-					$woo_extension    = $this->woo_extensions_list->get_woo_extension_slug_by_id( $woo_extension );
 					$woo_extension_id = $woo_extension;
+					$woo_extension    = $this->woo_extensions_list->get_woo_extension_slug_by_id( $woo_extension );
 				} else {
 					$woo_extension_id = $this->woo_extensions_list->get_woo_extension_id_by_slug( $woo_extension );
 				}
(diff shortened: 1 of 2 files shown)


MrJnrman:
Thanks for working on this @Luc45. It works great with absolute paths, but not with relative paths. Maybe we can pass the source through `realpath` when setting up the volumes here https://github.com/woocommerce/qit-cli/blob/9d76f3006ac58a4a908e790d58a46097e303b3ff/src/src/Environment/ExtensionDownload/ExtensionDownloader.php#L92

src/src/Commands/CustomTests/RunE2ECommand.php
```diff
@@ -273,6 +273,7 @@ protected function execute( InputInterface $input, OutputInterface $output ): in
 
 		putenv( 'QIT_UP_AND_TEST=1' ); // phpcs:ignore WordPress.PHP.DiscouragedPHPFunctions.runtime_configuration_putenv
 		putenv( sprintf( 'QIT_SUT=%s', $woo_extension ) ); // phpcs:ignore WordPress.PHP.DiscouragedPHPFunctions.runtime_configuration_putenv
+		putenv( sprintf( 'QIT_SUT_SOURCE=%s', $source ) ); // phpcs:ignore WordPress.PHP.DiscouragedPHPFunctions.runtime_configuration_putenv
```
Luc45, line 285: @MrJnrman what about we add realpath here?

MrJnrman: That'd do the trick

Luc45, line 276:
```suggestion
		putenv( sprintf( 'QIT_SUT_SOURCE=%s', realpath( $source ) ) ); // phpcs:ignore WordPress.PHP.DiscouragedPHPFunctions.runtime_configuration_putenv
```

MrJnrman (approved):
Works great with themes and plugins. Thanks Lucas! 🚢

PR #198, opened 2024-08-28
Add self-tests for the Validation test

Luc45:
This PR adds self-tests for the Validation test.

We added 4 scenarios:

- Main plugin
- Missing plugin
- Main theme
- Missing theme

Then we generated the snapshots of the results to make sure they are as expected.

### Testing Instructions
- Just review the snapshots and make sure they are as expected
- The ones with "Main" should be a success and the PHP header should be correctly identified
- The ones with "Missing" should be a warning, and the PHP header should be false

Examples:

image

image


singerb (approved):
Snapshots look correct, tests run correctly.

Luc45:
Sounds good, I'll be merging this when the upstream PR has been merged.

PR #200, opened 2024-09-02
Activation as a custom test

Luc45:
This PR does a couple of things to help support Activation Tests as a custom test. It's a counterpart of the work done in the Manager PR upstream.

- Automatically install dependencies of the SUT and additional plugins
  - Eg: Install direct dependencies of the SUT, or dependencies of additional extensions activated alongside the SUT.
- Also adds PHP extensions if the plugins requested any.
- Activate all plugins using the UI, dependencies first.
- Fix to allow to use both `--woo` parameter and a test tag of `woocommerce`, eg: `--woo nightly --plugin woocommerce:test:activation`, as the Activation Test has been published as a test tag of Woo.
- Added `RunActivationTestCommand` as a wrapper to running the test tag `woocommerce:test:activation`
- Automatically detect if the SUT is a plugin or a theme, and remove the `--testing_theme` flag
- On the Activation Tests, we use `--skip_activating_plugins` so that we can activate using the UI. However, if omitting this parameter, now we show the output of the WP CLI error when a plugin fails to activate using that method.
- Add a file inside the Playwright container on path `/qitHelpers/test-info.json`, which is a JSON containing the SUT type, slug and plugin activation stack (although the last one isn't being used)


MrJnrman (approved):
Tested this extensively locally and in staging and can confirm that everything is working as intended. Thank you for the great changes @Luc45

PR #203, opened 2024-09-06
Activation refactor backwards compatibility

Luc45:
This PR aims to minimize any backwards breakage, especially for CI scripts, on our refactor of the Activation Test Type.

It adds the following options to `run:activation`, which are often used in CI scripts: 

- `--json` - Outputs the JSON of the test at the end. The output is then usually piped to a file.
- `--wait` - Does nothing since the test runs in-loco now - it used to be async, and this would make it "sync".
- `--ignore-fail` - Exits with 0 even if the test fails - Can be useful for scripting, although I'd remove it if it weren't for backwards compatibility.
- `--zip` - Essentially an alias to `--source`

It's worth noting that I had to add all of these as actual options of the `run:activation` command. There is an [open PR](https://github.com/symfony/symfony/pull/54439) in Symfony that should be merged any time now that adds the capability for hidden options, would be super helpful here for things like aliasing `--zip` to `--source`, etc.

Even when this PR is merged, it will go into Symfony Console 7.3, which will require either PHP 8.2 or PHP 8.3, so to use it we would need to bump our minimum QIT CLI version to that version. We are currently on minimum of PHP 7.2.

All things considered, I think it's best to just keep listing these deprecated options. It's ugly, but bumping the minimum to PHP 8.2 just for this isn't worth it IMO.


Luc45:
Merging as discussed on Slack

PR #206, opened 2024-09-10
Update snapshots for validation tests

Luc45:
Update snapshots for validation tests per changes upstream.


singerb (approved):
Works for me; now that I fixed the bug the readme-only test works too (and I updated the snapshot there).

PR #207, opened 2024-09-18
Disable Mac test, add Validation workflow

Luc45:
The Mac workflow has been broken for a couple of days. 

It seems something in the GitHub pipeline broke the integration, they are looking into it: https://github.com/douglascamata/setup-docker-macos-action/issues/37

Disabling that workflow for now and removing it from the README.

Also adding Validation Workflow because it had self-tests but no Workflow.


zhongruige (approved):
LGTM!

PR #209, opened 2024-09-21
Add Tunnel support to enable sites to have a live URL and HTTPs

Luc45:
Some plugins might require a live site to be able to function properly, especially Payment Gateways and SaaS. Some of them might also require a HTTPS connection.

Our test environment runs locally and in CI. To accommodate this requirement, this PR adds support for ~Ngrok~ CloudFlare Tunnel, which is a reverse proxy that basically provides a live URL to the site running locally.

This PR has been refactored, switching from Ngrok tunnelling to CloudFlare Tunnelling, which requires no account, no setup at all, will "just work". Just pass `--tunnel` when spinning up an environment or running a test based on Custom Test, such as Activation or Custom E2E Test.

### Testing Instructions
- Checkout this branch
- Checkout the Manager counterpart branch (either locally or on Staging)

### Spinning up a local environment with a public URL
- `qit env:up --tunnel`

### Running an Activation Test with a public URL:
- `qit run:activation woocommerce-amazon-s3-storage --tunnel`

### Running a Custom E2E test with a public URL:
- `qit scaffold:e2e /tmp/foo`
- `qit run:e2e woocommerce-amazon-s3-storage /tmp/foo`

Assert in all scenarios, that the tests did in fact run using a public URL. You can do that running in verbose mode `-v` or by carefuly inspecting the Allure Report in Staging.


src/src/Tunnel/TunnelRunner.php
```diff
@@ -0,0 +1,82 @@
...
+	/** @var OutputInterface */
+	protected $output;
+
+	public function __construct( Docker $docker, OutputInterface $output ) {
+		$this->docker = $docker;
+		$this->output = $output;
+	}
+
+	public function start_tunnel( string $local_url, string $env_id ): string {
+		if ( ! empty( getenv( 'QIT_TUNNEL_REGION' ) ) ) {
```
rcstr: Wonder how can I define and test the region?

rcstr (commented):
@Luc45 I have tested your PR and can confirm tunnel is created, but I'm unable to get the tunneled site to work (is it expected?).

Screenshot 2024-09-30 at 06 19 44

If I visit https://heaven-dating-vt-easier.trycloudflare.com/. I get

Screenshot 2024-09-30 at 06 21 53

My docker-machine looks green also, 
Screenshot 2024-09-30 at 06 21 41

Let me know your thoughts.

src/src/Tunnel/TunnelRunner.php
Luc45: > Wonder how can I define and test the region?

This is kind of a hidden setting that I'm still wondering how to approach, it's currently an env var:

`QIT_TUNNEL_REGION=us qit env:up --tunnel`

As for the message above, in the screenshot, you actually shouldn't add that line to your hosts file for tunnelled connections, I've just pushed a commit to not show the "add to hosts file" message for tunnelled connections.

As to why it's not working for you I'd have to see. Do you have a few mins to spare for a screenshare session?

Luc45: @rcstr I realized that tunnelling through Docker won't work on Mac because it requires network mode set to `host`, which is only available in Linux. I've added a workaround, which is using a local tunnel that can be easily installed with one command. Can you please try it out again and see how it goes? I won't give too much detailed instructions as the CLI instructions should be enough - I think it's part of the review that the info in there is enough.

Luc45: PS: Just remember the checkout the Manager branch and point your CLI to local

zhongruige (approved):
LGTM! Thanks for this Lucas!

PR #212, opened 2024-10-04
Fix self-tests for the Custom Tests

Luc45:
PR description upstream.


zhongruige (approved):
Thanks for getting in these fixes @Luc45!

PR #213, opened 2024-10-04
Allow to use WooCommerce as the SUT for Activation Test

Luc45:
As per title, this should work now: `qit run:activation woocommerce --source https://github.com/woocommerce/woocommerce/releases/download/9.4.0-beta.2/woocommerce.zip`

Requires a small tweak to the upstream branch, which is a counterpart PR.


Luc45:
Merging this as per upstream comment.

PR #214, opened 2024-10-11
Prioritize extension slugs over local directories and improve Docker volume handling

Luc45:
The QIT workshop worked fine for most people, but at least 2 people had the following issues, which this PR fixes:

### Extension Slug Priority

Previously, running `qit run:e2e my-extension` in a directory containing a local `my-extension` folder would unintentionally pick up the local directory as the plugin to test. Now, it will only use the local directory when explicitly specified with `--source ./my-extension`.

### Docker Volume Paths

Ensures that volumes in the generated docker-compose use absolute or relative paths, preventing Docker from interpreting them as named volumes.

Example:

```yaml
volumes:
  - /home/john.doe/my-extension:/var/www/html/wp-content/plugins/my-extension
```

Instead of:

```yaml
volumes:
  - my-extension:/var/www/html/wp-content/plugins/my-extension
```


zhongruige:
Anything specific you'd like tested on this one @Luc45?

Luc45:
I wrote automated tests that covers the gist of these flows, but it's good to have a manual validation...

**1. Setup:**
- Scaffold a random test somewhere, like `qit scaffold:e2e /tmp/foo`
- Grab a copy of `woocommerce-amazon-s3-storage` as an example, extract it in a new directory
- Open a terminal in that directory

**2. Test remove takes precedence over local directory:** 
- Run a test with verbose mode, like `qit run:e2e woocommerce-amazon-s3-storage /tmp/foo -v`
- Take note of the plugin version, eg: 2.7.1
- Edit the plugin entrypoint and change the version to something else, like `2.7.1-local`
- Re-run and assert it still uses `2.7.1`

**3. Test you can use your local version if you want:**
- Now run it with `qit run:e2e woocommerce-amazon-s3-storage /tmp/foo --source woocommerce-amazon-s3-storage -v` and assert that it uses your local version. Feel free to try variations of source, like absolute paths or `./woocommerce-amazon-s3-storage`

zhongruige:
Forgot I had created this for the demo, minor improvement in the future we could consider is overriting it:

```
Directory already exists. Scaffold E2E tests in "/tmp/foo/" anyway? (y/n) y
Directory already exists: /tmp/foo/
```

Luc45:
Ahhh I see. About that, I could recursively delete the directory if the user answers `y`, or I could not ask anything and just bail with "Directory already exists".

My concern of recursively deleting the directory is if the user does something like `qit scaffold:e2e ./` in their home directory and confirm that recursive deletion by mistake...

zhongruige (approved):
This tested great for me, it picked up the version when the specific path was provided:

Screenshot 2024-10-11 at 1 52 47 PM

Nice work on this!

zhongruige:
> My concern of recursively deleting the directory is if the user does something like qit scaffold:e2e ./ in their home directory and confirm that recursive deletion by mistake...

Agree, I think we can throw an error and drop out rather than performing a destructive action just in case.

PR #215, opened 2024-10-11
Docker pull only once a day

Luc45:
There was one person at the workshop that hit a limit for Docker pulls.

Docker Hub allows 100 image pulls every six hours.

I suspect it's because we were all in the same network doing pulls together.

To minimize the risk of encountering this limit, this PR reduces the calls to `docker pull` to one per day, instead of per-request. This is faster, and reduces drastically the chances of encountering docker pull limits.

I'm not sure how to test this, you could add a breakpoint and check that the pulls happens once and then are early bailed because of cache hit, or you could try to check your current pull limits as per [this](https://www.docker.com/blog/checking-your-current-docker-pull-rate-limits-and-status/) article.


zhongruige (approved):
LGTM! Thanks for getting this in quickly.

PR #216, opened 2024-10-11
Overwrite scaffolded tests with safety checks

Luc45:
This PR fixes a bug with `qit scaffold:e2e <path>` when trying to scaffold a test on a directory that already exists, the user would be prompted to confirm that he wanted to overwrite the directory with the scaffold. But, upon replying `y`, it would bail anyway with "directory already exists".

The expected behavior would be to overwrite the target path with a newly scaffolded test, so in this PR, we delete the scaffolded directory if it already exists and scaffold a new test on the target directory.

However, there is a concern with deleting files on the user filesystem, in a scenario like this: `qit scaffold:e2e ./` on the home directory, which, if the user answered `y`, could potentially delete their home directory.

To prevent this scenario, this PR does a safety check, overwriting the directory only if it contains files we would expect to find in a scaffolded test. If there is any file that we do not expect, we do not delete anything.

### Testing Instructions
1. `qit scaffold:e2e /tmp/foo`
2. Run it again and confirm the overwrite, assert that it overwrites as expected
3. Change something in the files and scaffold again, assert that it overwrites as expected
4. Create a new directory, like `/tmp/foo/.config` and scaffold again, assert that it refuses to overwrite, as it contains unexpected files, and it asks the user to delete it themselves and re-running the command


zhongruige (approved):
Thanks @Luc45! This works great for parent directory files, such as `.env` or folders:

Screenshot 2024-10-11 at 2 27 51 PM

Screenshot 2024-10-11 at 2 28 22 PM

PR #219, opened 2024-11-12
Fix "docker compose not found"

Luc45:
When exploring which version of Docker Compose to use, we first try "docker-compose" then we try "docker compose".

This can produce a "docker-compose command not found" which is just informational and doesn't affect anything, but still, it's not good to display this to the user for no reason.

This PR changes this behavior so that Docker Compose V2 takes precedence, and also doesn't show that message anymore.

To test this, you can just make sure that things keeps working as usual. I've tested that both conditionals works locally.


zhongruige (approved):
Changes LGTM! Thanks for this tweak @Luc45!

PR #227, opened 2024-12-05
Fix Custom Tests Orchestration: Improve Plugin Actions, Dependencies, and Configuration Handling

Luc45:
This PR introduces a series of fixes for the Custom Tests, mainly around newly-released Orchestration.

- **Shared Setup/Teardown in `bootstrap` Mode:**  
  When additional plugins are set to `bootstrap`, their shared setup and teardown steps now run correctly, even if their test suite does not. This aligns with the intended behavior that `bootstrap` prepares the plugin environment but skips isolated tests.
- **Automatic `woocommerce` Dependency Handling:**
  If `woocommerce` is needed and not explicitly defined, it’s now added automatically when dependencies are requested. This respects the `--dependencies` parameter and streamlines setting up the test environment for WooCommerce-dependent plugins.
- **Enhanced Parameter Handling (`--tunnel`, `--dependencies`):**
  We now list all possible values for these parameters, improving discoverability. Additionally, if an invalid tunnel is provided, QIT will throw an error rather than silently defaulting to `auto`.
- **Download Type `tests` Only:**
  Introduced an optional `download_type=tests` property, ensuring that only the test artifacts for the plugins under test are downloaded. This is particularly useful in staging environments that may not have access to all plugin zips, especially when a local source is provided.
- **Action Overrides for Additional Plugins:**
  Previously, when a plugin’s `action` was defined in `qit.yml`, attempting to override it at runtime using `-p plugin:action` or `--plugin plugin:action` did not always produce the intended results. Now, we explicitly apply CLI overrides **after** loading `qit.yml` and processing dependencies. This ensures that runtime parameters correctly override `qit.yml` settings. For example, you can switch `woocommerce-progressive-discounts` from `test` to `bootstrap` or `activate` at runtime, and the final environment configuration, tests, and reports reflect this change correctly.
- **Refined Configuration Loading Order:**
  We’ve adjusted when `qit.yml` is loaded during the environment lifecycle. Previously, the SUT’s `source` could inadvertently revert to its slug, ignoring a local directory defined in `qit.yml`. Now, `qit.yml` settings (including `source`) are respected and not overwritten by defaults.

- **Comprehensive Tests and Snapshots:**
  We added extensive scenario-based tests and snapshot testing to `PlaywrightOrchestration` and `RunE2ECommand`. These tests confirm that different plugin actions (`test`, `bootstrap`, `activate`), CLI overrides, dependencies, and tunneling options produce the expected final configuration and result in the appropriate setup, teardown, and test steps being run.

- **Fix Orchestration Self-Test Workflow:**  
  Resolved issues in the orchestration self-test workflow to ensure it accurately verifies the test environment setup. This fix prevents false positives and ensures that the self-test reliably reflects the actual state of the orchestration, enhancing the overall reliability of the testing process.

- **Fix typo in dependencies.json:**
  Changed the path for loading "qit/dependencies.json" to "bootstrap/dependencies.json", which was a typo.
  
- **Break down `RunE2ECommand` into smaller classes:**
  To make it easier to maintain and reuse this code in the future (for example, when adding additional local test commands like `RunSecurityCommand`), we reorganized the previously monolithic `RunE2ECommand` class into a set of smaller, dedicated classes:
    - **AbstractRunTestCommand:** Provides a common foundation for running local tests, enabling future commands to reuse and build upon the same logic.
  - **ConfigurationProcessor:** Centralizes and simplifies the handling of CLI arguments and `qit.yml` settings.
  - **EnvironmentRunner:** Manages environment setup and teardown.
  - **RunE2ECommand:** Now leverages these helper classes, resulting in a more maintainable codebase that is easier to understand and extend.

**Testing Instructions:**

- Pull this branch
- Deploy the counterpart Manager branch to Staging

- **Basic Run (No Overrides):**  
  From `_tests/orchestration`, run:  
  `qit run:e2e woocommerce-amazon-s3-storage`  
  Observe that it completes successfully, and both the SUT and additional plugins (like `woocommerce-progressive-discounts`) run their full test suites.

- **Changing Actions via `qit.yml` or CLI:**  
  - Set `woocommerce-progressive-discounts` to `bootstrap` in `qit.yml` and re-run. Confirm that shared setup and teardown run, but the plugin’s main tests and isolated setups/teardowns do not.
  
  - Alternatively, run:  
    `qit run:e2e woocommerce-amazon-s3-storage -p woocommerce-progressive-discounts:bootstrap`  
    This runtime parameter achieves the same effect as editing `qit.yml`, allowing you to override the plugin’s action on the fly.

- **Activate Mode and Dependencies:**  
  Switch `woocommerce-progressive-discounts` to `activate` mode (via `qit.yml` or CLI) and confirm that it is activated but does not run tests or setup/teardown steps. Adjusting `_tests/orchestration/woocommerce-amazon-s3-storage-tests/bootstrap/setup.js` to remove references to it will confirm it’s present but not tested.

Check that all self-tests continue to pass on Staging, including Activation.

**Example Outputs:**

- **`woocommerce-progressive-discounts` in `test` mode:** Runs full tests.
image
- **`woocommerce-progressive-discounts` in `bootstrap` mode:** Shared setup/teardown only, no isolated tests.
image
- **`woocommerce-progressive-discounts` in `activate` mode:** Simply activated, no tests or setups run.
image


zhongruige (approved):
This tested great for me, I was able to kick off a series of tests (including Woo API and Woo e2e just to check) and everything worked as expected for me. Nice work on this!

PR #228, opened 2024-12-11
Fix Woo Extension ID-to-Slug Conversion Bug and Refactor RunE2ECommand for Readability

Luc45:
## What this PR does

**Fix:** Fix a bug that was introduced in yesterday's release 0.8 when passing an extension ID instead of a slug, eg:

When running `qit run:activation 2165910`, it would give this error:

`The extracted zip '/home/runner/work/woocommerce-shipping/woocommerce-shipping/woocommerce-shipping.zip' file does not contain a parent directory matching the slug '2165910'.`

Now it converts the ID to the slug as expected before moving on.

**Cleanup:**
- Removed AbstractRunTestCommand
- Broken down RunE2ETestCommand's `execute` method into smaller, more readable methods

**Tests:**
- Added tests to cover the scenario of a Woo Extension ID being passed

## Reviewing this PR

This PR consists of 4 commits, I recommend reviewing this PR by looking at it's commit history, instead of Files Changed:

1. First, I added a test that replicates the bug, in [this commit](https://github.com/woocommerce/qit-cli/commit/587a9910eea9f1dfe65a1cc62be6b8170045693e). We can see that the `slug` is numeric when given a Woo Extension ID as input.
2. Then, [I fixed the bug](https://github.com/woocommerce/qit-cli/commit/8bb222cfb59b836fabeb30cbd4f30734637299cb), which we can verify by the updated snapshot in this commit
3. Next I [refactored the fix](https://github.com/woocommerce/qit-cli/commit/f974ace3953a4ebb54fa38c1ae19f785a8d94a31) to keep the implementation as close as possible to the original, asserting that it still fixes the bug, by making sure that the snapshot is unaffected.
4. Lastly, in the [fourth commit](https://github.com/woocommerce/qit-cli/commit/abed640ae483b849b2db6057b67cf802d51e62ff#diff-ce44bde6248bb2a67ff3155bea4a2c02fc54362b252450106a0e54574b091587R168) I refactored `RunE2ECommand` just to break down the `execute` method into smaller methods to make it easier to read. The `execute` method had 230 lines before, and after it was broken down into smaller methods it ended up with 166 lines, while preserving the exact same functionality as before.


zhongruige (commented):
Kicked of a few activation and woo-e2e tests using the binary on this PR and they worked as expected, using a combo of slugs and WooCommerce IDs.

@Luc45 I did notice this when using the plugins param:

```
./qit run:activation automatewoo --plugin=2165910
Downloading plugins and themes...
No download URL found for 2165910
```

I was able to [run a test](https://qit.woo.com/?qit_results=4709260.aeK04t5jOYEjMSWpiO5rBecQcs2IShta5zbxNwV0QeHaYnDTKvWo5tB79deH2l6X) for this:

```
./qit run:woo-e2e 2165910 
```

I believe this should also accept IDs, so wanted to double check if it's something we need to change here (or a follow up PR is also fine to get this one out).

zhongruige:
Noting that `./qit run:activation automatewoo --plugin=automatewoo-birthdays` does work as expected.

Luc45:
Nice catch, on it.

Luc45:
@zhongruige Can you please test again? Should be working with `--plugin` now
image

Luc45:
Similar to before, first I added a test to reproduce the bug, which we can see in the snapshot here:

image

Then I fixed the bug, and validated that the snapshot was as expected upon updating it:

image

zhongruige (approved):
Thanks for getting in this fix @Luc45! Tested great for me, happy to add my approval.

PR #229, opened 2024-12-12
Fix activation test and dependencies

Luc45:
## Enhanced Handling of the Activation Test’s SUT Actions

Our [Activation Test](https://qit.woo.com/docs/managed-tests/activation) is, under the hood, a custom test of the `woocommerce` under the `activation` test tag.

So to run it, we ultimately do something like this:

```bash
qit run:e2e my-extension -p woocommerce:test:activation
```

So we are testing `my-extension` as the SUT, with the `activation` test tag of `woocommerce`.

Since the SUT always have the `test` action, it was trying to run the test phase of the SUT on the Activation tests as well.

Usually this is okay, but this week we had a case where the custom test of an extension was failing, and that was affecting the `activation` test.

This PR fixes this by making sure that the SUT has a `bootstrap` action, and I also took the opportunity to make it run the `pre-activation` test tag, to give a way for the plugin developer to run something before the activation test, in case they want to get rid of some kind of onboarding wizard that might be affecting the activation test.

If the extension doesn't have a `pre-activation` test tag, it will just run the Activation Test normally.

## Corrected Inclusion of Plugin Dependencies

As part of this PR, I also saw that the dependencies of plugins were not being included correctly due to a small logic typo introduced in 0.8. This fix can be observed in [this snapshot](src/tests/Environment/__snapshots__/RunE2ECommandTest__test_with_plugin_dependencies_cli_only__1.json) for instance, which now adds the plugins that are listed as dependency of either the SUT or an additional extension requested in the test. In the Manager for a while now to assert that, when requested, we also return dependencies of dependencies if needed.

Example: `qit run:activation automatewoo-birthdays`

- `AutomateWoo Birthdays` depends on `AutomateWoo`
- `AutomateWoo` depends on `WooCommerce`

These plugins should be active in the environment:
- WooCommerce
- AutomateWoo
- AutomateWoo Birthdays

Fix fix also applies to any extra PHP extensions needed by any plugin in the stack.


zhongruige (approved):
Changes look and tested good for me, super strange GitHub is rendering the comments weird as if they aren't comments, but I double checked the files in my IDE and they seem fine so 🤷 

Screenshot 2024-12-13 at 11 22 16 AM

PR #231, opened 2024-12-23
Add command to get multiple test results in one request

Luc45:
This command is used by the new self-test script, which is used in the 9.6.0-beta.1 sync.

It allows to get multiple tests with one command.

### Testing instructions
- Get a couple of test run IDs with `qit list-tests`
- Get multiple tests in one command with `qit get-multiple "123, 456, 789`
- Add `--json` and see that the response is a JSON arrary where each item is each test


zhongruige:
Overall looks good to me @Luc45, I did notice some large spacing for activation tests, are you seeing the same?

JSON formatting looks good.

Luc45:
Fixed, eg:

image

zhongruige:
Ah awesome thanks @Luc45!

zhongruige (approved):

PR #232, opened 2024-12-23
Self-test script rewrite

Luc45:
### Background & Motivation
The motivation for this PR came after I wasted a couple of hours while doing the sync with 9.6.0-beta.1. One of the self-tests didn't start, and I only found out about this 2h later. Debugging why it failed was much harder than it should be, because the self-test script would just dispatch 20 tests in parallel, keep them all alive with `--wait` and `--json`, where each would poll their own test status individually.

This PR greatly simplifies this setup, we dispatch the tests async and collect their test run IDs, then we poll their status ourselves with one request to the newly introduced `get-multiple` endpoint, which takes a comma-separated list of test run IDs.

This made it possible to also improve the output, and added extensive logging to all requests to `last-self-test.log`, that gets reset with each self test. 

#### **Before**
- **Multiple Parallel Processes**: We would start many processes, each:
  1. Dispatching its QIT test.
  2. Polling individually for status until completion.
  3. Reporting success or failure.

- **Drawbacks**:
  - Hard to coordinate and debug: each process produced its own logs and status updates.
  - Risk of wasted time: if one test fails to run, it would be very hard to know why.
  - More overhead managing N distinct polling loops concurrently.

#### **After**
1. **Asynchronous Dispatch**  
   - All tests are **dispatched quickly** (no blocking).  
   - We gather every `test_run_id` upfront.

2. **Unified Polling Loop**  
   - Rather than multiple pollers, **a single loop** calls `qit get-multiple` for all `test_run_id`s in one go.  
   - This repeats every X seconds until each test reports a final status.

3. **Snapshot Verification**  
   - As soon as a test finishes, its JSON result is stored and checked against the PHPUnit snapshot.  
   - Any mismatch shows up right away (or updates, if in `update` mode).

### Implementation Details
- **`QitRunner.php`**: 
  - Dispatches all tests asynchronously and records their IDs. 
  - Repeatedly calls `qit get-multiple` until all have a final result.
- **`QITLiveOutput.php`**:  
  - Adjusted to show a unified table of test statuses during each poll iteration.
- **`PhpUnitRunner.php`**: 
  - Receives completed test data and checks it against snapshots.
- **Documentation**:  
  - Updated to explain the single polling flow (instead of many parallel pollers).

### Testing instructions
- Run the self-tests as usual, see how it goes


zhongruige:
I'll dig into this a bit more tomorrow and run the tests but so far loving these changes!

Luc45:
No worries! Ignore if you see some activity in GH, I'm just taking the opportunity to run some self-tests and scheduling the PC to auto-shutdown in a couple of hours.

_tests/managed_tests/README.md
```diff
@@ -1,42 +1,176 @@
...
-- Each directory is a test-type self-test, except for "vendor" and "tests" folders, eg: `activation`, `security`, `e2e`, etc.
-- Each test-type self-test can contain multiple tests, eg: `activation/test-1`, `activation/test-2`, etc.
-- Each test has a `env.php` file, with parameters used for the test, such as PHP, WooCommerce and WordPress versions, eg: `activation/test-1/env.php`.
+1. **Discovery & Packaging**
+  - Each **test type** (e.g., `activation`, `security`, `woo-e2e`) has one or more **test scenarios** (e.g., `test-1`, `scenario-a`).
+  - The script packages each scenario’s System Under Test (SUT) into a ZIP (unless reusing JSON snapshots).
 
-### Running the self-tests
+2. **Dispatch**
+  - The script uses `/home/lucas/automattic/qit-cli/qit` to **run** each scenario test asynchronously, collecting `test_run_id`s.
```
zhongruige: I think we should remove the dir path here and below, so it's like this:

```suggestion
  - The script uses `qit-cli/qit` to **run** each scenario test asynchronously, collecting `test_run_id`s.
```

I might even reword this to something like:

```
  - The script uses the binary at `qit-cli/qit` to **run** each scenario test asynchronously, collecting `test_run_id`s.
```

_tests/managed_tests/README.md
```diff
@@ -1,42 +1,176 @@
...
 
-## Running the self-tests
+---
 
-- `./vendor/bin/phpunit ./tests/self_test`
-- Or `./vendor/bin/phpunit -d --update-snapshots ./tests/self_test`
\ No newline at end of file
+## Running Tests
+
+1. **Navigate** to `/home/lucas/automattic/qit-cli/_tests/managed_tests`.
```
zhongruige: Directory change recommendation:

```suggestion
1. **Navigate** to `_tests/managed_tests`.
```

zhongruige: I won't comment on all of the instances here, but I would recommend we just use the `qit-li` as the current working directory and drop `/home/lucas/automattic/` as it could be confusing if they need that directory structure or not.

zhongruige:
Went through the self tests (`php QITSelfTests.php`) and it batched and incremented as it was running:

Screenshot 2024-12-24 at 8 39 49 AM

However, I did hit a timeout and it eventually exited:

Screenshot 2024-12-24 at 12 15 32 PM

Wanted to just double check on the above behavior.

Luc45:
> Wanted to just double check on the above behavior.

Nice catch, I've limitted the `get-multiple` endpoint to 20 tests per request on the backend, and updated the logic here to:

- Stop requesting tests that are complete
- Batch polling in chunks of 20 tests, instead of trying to fetch them all at once

I was able to run a mass test with all tests at once, so it seems to have solved it.

_tests/managed_tests/README.md
Luc45: Done, thanks!

zhongruige:
Re-ran this again and just finished (ran this twice just to see) and got this result both times:

Screenshot 2024-12-27 at 9 00 14 AM

Would we expect these all to pass or is this expected as we have some no-op tests we might expect to fail?

Luc45:
It should pass on the 9.6.0-beta.1 sync branch

zhongruige:
Weird I got the same results on that branch @Luc45:

```
All snapshots have been verified.

Some snapshots still failed. Final outcome: ❌

For more details, see last-self-test.log.
➜  managed_tests git:(24-12/sync-960) 
```

Not sure, could it be due to the PHP version I'm using?

zhongruige:
Perfect that fixed it, thanks @Luc45!

Screenshot 2024-12-27 at 12 47 19 PM

zhongruige (approved):

PR #235, opened 2025-01-08
9.6.0-beta.2 sync

Luc45:
No snapshot changes, except for debug log changes in `delete_products` test that we always ignore. Since we always ignore it, this PR makes no distinction between debug log in these tests being empty or having something and being ignored.


zhongruige:
> this PR makes no distinction between debug log in these tests being empty or having something and being ignored

Love this! This is a great improvement!

zhongruige (approved):

PR #236, opened 2025-01-10
Enhance `extract_error_info` Function to Support Additional Error Formats

zhongruige:
This update enhances the extract_error_info function to support additional PHP error formats, including PHP Parse error and uncaught exceptions. The function now captures a wider range of error types to improve error parsing and reporting.


zhongruige:
Closing, we'll go with the changes @Luc45 has prepared including tests around this.

PR #237, opened 2025-01-10
Detect parse errors

Luc45:
Also detects PHP Parse Error as a type of fatal error, per Slack convo.


src/tests/LocalTests/PrepareQMLogTest.php
```diff
@@ -0,0 +1,23 @@
...
+use Spatie\Snapshots\MatchesSnapshots;
+
+class PrepareQMLogTest extends QITTestCase {
+	use MatchesSnapshots;
+
+	public function test_parses_fatal_and_parse_errors() {
+		$debug_log = <<<'TXT'
+[10-Jan-2025 17:12:43 UTC] PHP Parse error:  syntax error, unexpected '|', expecting variable (T_VARIABLE) in /var/www/html/wp-content/plugins/kymvex-inv-alert/vendor/phpunit/phpunit/src/Framework/Assert/Functions.php on line 83
+[10-Jan-2025 17:12:43 UTC] PHP Fatal error:  Exception thrown without a stack frame in Unknown on line 0
+[10-Jan-2025 17:52:22 UTC] PHP Fatal error:  Uncaught Exception: Foo in /var/www/html/wp-content/plugins/kymvex-inv-alert/kymvex-inv-alert.php:29
```
zhongruige: Only nit would be to use a generic fake slug for the product instead of this one, for here and the snapshot

zhongruige (approved):
Overall looks good @Luc45, just a recommendation to change the slug used in the tests

src/tests/LocalTests/PrepareQMLogTest.php
Luc45: Nice catch! Thanks.

PR #243, opened 2025-01-30
Update mocked schema for unit tests 

Luc45:
**PS: The base of this PR is your extension set PR, @singerb , so consider this PR a code review suggestion that, if merged, will be part of your extension set PR.**

The unit tests were failing because `extension_set` wasn't found in the Schema used by the tests. The schema was mocked in `sync.json` for the purpose of the tests.

The response was mocked as it contained dynamic data such as `wordpress_version` enums, so using a static mock allowed to have a predictable response for snapshot testing.

This PR changes this behavior by fetching the `sync.json` file fresh from the Manager backend (as the CLI would) on every PHPUnit test run, and normalizing it before saving. This fixes the `extension_set` not found bug.

This PR also improves the `DynamicCommandCreator` to throw an exception if a requested parameter isn't found in the schema, to make this kind of error clearer in the future.

It also fixes Xdebug in unit tests, etc, which for some surprising reason wasn't implemented. Maybe it was silently removed in some recent Makefile changes by mistake, but I didn't dug into it to see history, etc, or maybe it was never implemented in this repo at all. Either way, it's in place now, and running `make phpunit DEBUG=1` will have Xdebug working as expected.

The only thing is that the Schema is being downloaded from Manager Prod. I thought about changing this to Manager Staging. Alternatively, instead of updating `sync.json` on every test run, we could instead allow the developer to update it by running the `pull-sync-json.php` script directly, and have him choose the environment to get the sync response from.

This way the developer could use a sync response from his own environment and commit that to the repo as the new "mocked response". My only concern is that the manual update adds a bit of complexity, one extra step. Thoughts?


singerb (approved):
Yeah, this looks reasonable. In general I think running the sync on test and hitting production are both reasonable choices; my one concern would be that you have to merge and deploy any QIT-side changes before your CLI changes can pass tests, but this isn't likely to be an issue often outside of say, this pair of PRs. And hitting staging doesn't change that anyway.

PR #245, opened 2025-02-15
PHPStan Snapshots

Luc45:
This PR has the updated snapshots for the PHPStan Test Type improvements.

Detailed description upstream.


zhongruige (approved):
Thanks for updating the snapshots!

PR #247, opened 2025-02-18
Adds ignoring to the log parsing for some common deprecation warnings

singerb:
When testing the activation test with the `compatibility` extension set, I discovered some deprecation warnings that were coming from the set members, rather than the SUT:

```
Class Jetpack_Geo_Location is deprecated since version 14.3 with no alternative available.
&
Hook woocommerce_get_price is deprecated since version 3.0.0! Use woocommerce_product_get_price instead.
```

The first is coming from Jetpack itself, since it adds geo-location support to the 20xx themes, but has also deprecated its geo-location support. This is indeed a little silly, and Jetpack is updating that behavior. Supressing this in the meantime is good.

The second comes from some common code in Subscriptions and Payments. It'd be nice to address, but is also probably low priority. Also a reasonable candidate to ignore - as long as we're not testing one of those two, in which case we should surface it, since they could do something about it (vs some random SUT just using the extension set).

Testing: this one is tricky. The activation test actually uses the Query Monitor logs, and a separate change is coming to filter these messages there. However, the Woo API and Woo E2E test types can also use extension sets, and do use the debug log we're parsing here, so fixing it for them is reasonable. Those run in GH actions though, so I'm less sure how to get the new QIT binary in there to test it. I've tested it locally with the activation test and some logging, and confirmed that it does skip the desired lines, so maybe that's good enough?

This will need a new release once merged, I believe.

### DIFF
--- a/src/src/LocalTests/PrepareDebugLog.php
+++ b/src/src/LocalTests/PrepareDebugLog.php
@@ -68,7 +68,7 @@ class PrepareDebugLog {
 					continue;
 				}
 
-				// Some compatbility extension set issues:
+				// Some compatbility extension set issues.
 
 				// Ignore a deprecation warning from Jetpack about itself.
 				$is_jetpack_geo_deprecation = stripos( $line, 'Class Jetpack_Geo_Location is' ) !== false;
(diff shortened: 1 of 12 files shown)


Luc45:
@singerb I have to do some normalization to these snapshots. You can pretty much ignore these PHP errors for now, please.

singerb:
> @singerb I have to do some normalization to these snapshots. You can pretty much ignore these PHP errors for now, please.

Cool. I fixed (hopefully) the PHPCS issues and merged trunk in just in case.

zhongruige:
@singerb looks like we see the `Jetpack_Geo_Location` same logs on Jetpack (example: https://qit.woo.com/?qit_results=4747736.B81NMkXO1goJDqhLJmdURcPF2o21oViZpCDi1JQpxyu0ofLwJ8oY3IX58WagvJag) so we can run a test on it after this is deployed to validate the changes

zhongruige (approved):
Approving as the changes seem straightforward!

Luc45:
@singerb I've just pushed the normalizations of those PHPUnit snapshots in this PR

PR #248, opened 2025-02-18
Allow to use integer zero as value for required option built from schema

Luc45:
Fixes `qit run:phpstan foo --phpstan_level 0`, which previously complained that `--phpstan_level` was required.


zhongruige (approved):

Luc45:
This is failing for the same snapshot changes fixed on the other PR. I'll just merge.

PR #249, opened 2025-02-19
Add Playwright test tag support

MrJnrman:
Add PW test tag support for custom and activation tests.

### Testing Instructions
1. Checkout this branch 
2. Run an activation test with a test tag using `./qit run:activation <slug> --pw_test_tag basic`
3. Confirm that it runs only the tests tagged as `basic`
4. Repeat step 2 without the `pw_test_tag` param.
5. Confirm it runs all tests.
6. Run any custom test with the `--pw_test_tag basic` tag and confirm that it doesn't run any tests.


Luc45:
You could also use `--pw_options="--grep @basic"` for the same effect. Basically, any string in `--pw_options` is passed directly as a parameter to Playwright bin.

zhongruige (approved):
Tested great for me, thanks for working on this!

PR #250, opened 2025-02-21
Filter deprecated query monitor logs in the CLI too

singerb:
Per the discussion in https://github.com/Automattic/compatibility-dashboard/pull/1231, this PR:
- Passes the env info down so that `PrepareQMLog` knows what the SUT is.
- Filters out the deprecation warnings we care about in `PrepareQMLog`. This has more context available than doing the filtering in the mu-plugin itself inside the environment.

Testing:
- Use https://github.com/Automattic/compatibility-dashboard/pull/1231 and trust that I copied the right binary into it.
- Run an activation test with the `compatibility` set, following the instructions in that PR.
- Verify there are still no deprecation issues.

Once this is merged I'll do another CLI release, and then copy that proper version into the manager code and update the linked PR.


singerb:
> Once this is merged I'll do another CLI release, and then copy that proper version into the manager code and update the linked PR.

@MrJnrman I assume we've been releasing the CLI when we do changes that we include the activation test? I was thinking maybe we can skip it this time, but I guess this is also useful for people running activation tests locally, so it's worth releasing.

src/src/LocalTests/PrepareQMLog.php
```diff
@@ -179,6 +181,25 @@ public function summarize_qm_logs( string $directory ): array {
 		foreach ( $data as $hash => $logs ) {
 			foreach ( $logs as $type => $type_logs ) {
 				foreach ( $type_logs as $info ) {
+					// Ignore some compatbility extension set issues.
+
+					// Ignore a deprecation warning from Jetpack about itself.
+					$is_jetpack_geo_deprecation = stripos( $info['message'], 'Class Jetpack_Geo_Location is' ) !== false;
```
MrJnrman, line 187: Do we want to ignore this if we are testing jetpack?

MrJnrman:
> @MrJnrman I assume we've been releasing the CLI when we do changes that we include the activation test? I was thinking maybe we can skip it this time, but I guess this is also useful for people running activation tests locally, so it's worth releasing.

Agreed on releasing. I also have [PR](https://github.com/woocommerce/qit-cli/pull/249) that makes some modifications to activation and custom tests so we can bundle them in the same release.

src/src/LocalTests/PrepareQMLog.php
singerb: IMO yes, they're doing it to themselves. Also it'll get removed at some point, depending on what release the PR makes it into.

singerb:
> > @MrJnrman I assume we've been releasing the CLI when we do changes that we include the activation test? I was thinking maybe we can skip it this time, but I guess this is also useful for people running activation tests locally, so it's worth releasing.
> 
> Agreed on releasing. I also have [PR](https://github.com/woocommerce/qit-cli/pull/249) that makes some modifications to activation and custom tests so we can bundle them in the same release.

Cool, we can coordinate that.

singerb:
Updated after more testing to only do the Jetpack filtering. Will update the other PR with the full release once done.

singerb:
Unit test failures don't appear related, and it's up to date with trunk.

singerb:
@MrJnrman Snapshots are fixed, this should be good to go?

MrJnrman (approved):
Changes look great. Thanks @singerb

MrJnrman:
@singerb Good to go. Added my approval

PR #251, opened 2025-02-24
Update snapshots for PHPStan level

singerb:
Unit tests are currently failing on trunk since these snapshots were missing the phpstan level parameter.


Luc45 (approved):
Thanks!

PR #252, opened 2025-02-25
Add key to snapshots

Luc45:
Add a script to add a key to the snapshots without having to re-run all the tests. Uses it to add `test_variation` after `phpstan_level` as expected.


zhongruige (approved):
LGTM!

PR #256, opened 2025-03-11
Improve `qit get` command by removing some long columns

Luc45:
- Hides some big/irrelevant columns when running `qit get <test_run_id>`: `ctrf_json`, `debug_log`, `test_type`, `runner`, and `workflow_id`.
- Removes the column that was set to 80, which can break links


src/src/Commands/GetCommand.php
```diff
@@ -142,13 +157,12 @@ protected function execute( InputInterface $input, OutputInterface $output ): in
...
 
 		$table = new Table( $output );
-		$table->setColumnMaxWidth( 1, 80 );
 		$table
 			->setHorizontal()
 			->setStyle( 'compact' )
 			->setHeaders( array_keys( $test_run ) )
 			->setRows( [ $test_run ] );
-		$table->render();
+	$table->render();
```
zhongruige: Not sure if this was intention but looks like a tab was removed here

Luc45: :+1: PHPCBF to the rescue

zhongruige (approved):
Tested great, current version on top and the changes on this PR on the bottom:

Screenshot 2025-03-11 at 10 30 20 AM

Also made sure no impact on the existing table.

Luc45:
> Tested great, current version on top and the changes on this PR on the bottom:
> Screenshot 2025-03-11 at 10 30 20 AM
> 
> Also made sure no impact on the existing table.

Yup, the real change is on test types that have the `ctrf_json` property, such as `woo-api`, `woo-e2e`, `activation` or `e2e`.

PR #257, opened 2025-03-12
Fix WPORG download slug ambiguity

Luc45:
This PR fixes an issue on `qit env:up` when trying to download an extension from WordPress.org that exists as both a slug and a theme. Details on upstream PR.


zhongruige (approved):
Changes tested great for me, thank you @Luc45!

PR #259, opened 2025-03-13
Activate theme with parent

Luc45:
When activating a theme, take into account the scenario of child-parent theme relationship.

- Just one --theme: Activate it
  - If it requires a parent theme, try installing parent from WPORG first, otherwise ask user to provide it
- Two themes: Check if one is parent and another is child, activate the child
- Two unrelated themes: Skip activating
- Three or more themes: Skip activating


zhongruige (approved):
This tested great for me, thank you @Luc45!

zhongruige:
Already approved this but went through and tested the latest changes which worked great for me, thanks so much @Luc45!

PR #265, opened 2025-03-18
Tolerate missing docker

Luc45:
Allows to use the QIT CLI without Docker.

Todo: Early bail if interacting with `run:e2e` or `run:activation` if Docker is not installed.


zhongruige:
> Todo: Early bail if interacting with run:e2e or run:activation if Docker is not installed.

Do we want to add in a follow up issue for this on to tackle in the future @Luc45?

Luc45:
I can that in in this PR

zhongruige:
Ah! No worries, whatever is easiest!

src/src/Environment/Docker.php
```diff
@@ -24,12 +24,12 @@ public function __construct( OutputInterface $output, Cache $cache ) {
...
 		$docker = 'docker';
 
-		$docker_version = trim( shell_exec( $docker . ' --version' ) ); // phpcs:ignore WordPress.PHP.DiscouragedPHPFunctions.system_calls_shell_exec
+		$docker_version = trim( shell_exec( $docker . ' --version 2>/dev/null' ) ?? '' ); // phpcs:ignore WordPress.PHP.DiscouragedPHPFunctions.system_calls_shell_exec
 
 		if ( $docker_version ) {
 			return $docker;
 		} else {
-			throw new \RuntimeException( 'Could not find docker' );
+			throw new \RuntimeException( 'It seems Docker is not installed. Docker is needed to run some test types.' );
```
zhongruige: Minor nit but I think we can direct them to the Docker docs here (doesn't need to be an explicit URL) but something like:

```suggestion
			throw new \RuntimeException( 'It seems Docker is not installed. Docker is needed to run some test types. For more information, please see the official documentation on installing the Docker engine for your operating system.' );
```

src/src/Environment/EnvironmentDanglingCleanup.php
```diff
@@ -70,6 +70,16 @@ protected function debug_output( string $message ): void {
 	}
 
 	public function cleanup_dangling(): void {
+		try {
+			App::make( Docker::class )->find_docker();
+		} catch ( \Exception $e ) {
+			if ( $this->output->isVerbose() ) {
```
zhongruige, line 76: Thanks for wrapping this behind the verbose flag!

zhongruige (approved):
This piece tested great for me, thanks for adding this in, especially in response to the Zendesk ticket!

I did leave a comment suggested a minor wording change, so let me know what you think @Luc45!

PR #267, opened 2025-03-18
Set correct SUT path for zip file

Luc45:
`qit.json` inside `automatewoo.zip` should work now.

There's an unrelated `wp-cli.yml` change in this PR that was part of another branch, but I'm adding as part of this PR because I forgot to push it earlier.


zhongruige:
Just tested this with the same zip file but didn't seem to get picked up in the test report, using the same `./qit run:activation <extension-slug> --zip` command

zhongruige (approved):
Tested this one again, used the zip that didn't have the qit.json file in it (the one I was using to benchmark the behavior) but it's working now with the correct zip file! Thanks for triaging this with me @Luc45!

PR #270, opened 2025-03-20
Fix latest version requirement and update notification

Luc45:
This PR fixes the update check and the latest-version enforcement check, preventing it from early bailing.

What was happening is that in the source code, we have this:

```php
// Do not check version on development build.
if ( $current_version === '@QIT_CLI_VERSION@' ) {
	return;
}
```

However, during the build step of the phar, that becomes:

```php
// Do not check version on development build.
if ( $current_version === '0.8.6' ) {
	return;
}
```

Which effectively early bails the update check and the latest version enforcement logic.

This PR fixes this by comparing the version to a base64 encoded representation of the development version placeholder, so that ultimately, it does this check at runtime:

```php
// Do not check version on development build.
if ( base64_encode( $current_version ) === 'base64-encoded representation of @QIT_CLI_VERSION@' ) {
	return;
}
```

Essentially preventing `@QIT_CLI_VERSION@` from being replaced with `0.8.6` or whathever version placeholder there was, since it's now base64 encoded.

### Testing Instructions

- Pull the branch from the Manager that sets `0.8.7` as the minimum QIT CLI version allowed
- Pull this branch and do `make build VERSION=0.8.6`
- Try to interact with your local Manager

See that you get a **forced updated prompt**:

image

Now change the endpoint in the Manager to require 0.8.6 as the minimum (leaving the latest unchanged) and repeat. See that you get a soft update promp, and that the original command completes:

image


zhongruige (approved):
Tested great for me!

PR #271, opened 2025-03-24
Remove working directory awareness

Luc45:
Makes QIT more explicit by removing working directory awareness, which was a feature that attempted to map the current plugin automatically if running QIT from inside a plugin directory as a QoL. But might not be the intent of the user, and we are trying to make QIT more explicit rather than magical to keep it simple, expressive, and reduce confusion.


zhongruige (approved):
Changes tested great for me, tried a variety of tests and scaffolded e2e tests to ensure  everything still worked as expected.

PR #272, opened 2025-03-26
Cache undefined versions for 1h instead of 1m

Luc45:
WPORG is getting sensitive when running multiple tests in sequence. This PR increases how long we cache undefined versions from 1 minute to 1h.

The 1 minute cache was meant to just avoid being throttled, because we essentially didn't have a reliable way to cache burst since we didn't have a version.

This PR increases how long we cache in this scenario to 1h, which is the intended behavior on 99.9% of the cases.


zhongruige (approved):
Changes LGTM!

PR #261, opened 2025-03-14
Add Group Test Support

MrJnrman:
### Description

This PR add support for grouped and batched tests. A group is a set of user defined tests which are registered as a logical and queryable uni, while a batch is similar except that the tests have no form of grouping. 

### Changes in this PR
- Adds a new `--group` option for `run:<test>` commands that specify that the test run shouldn't be immediately executed, but instead be added to a group to be executed later.
- Adds a new `group:register` command that is used to register but not execute test groups in the QIT manager. It takes the following options:
   - `group-identifier` - a user specified group identifier for easy querying
- Add a new `group:run` command that executes group of tests. They take takes the following options:
   - `group-identifier` - a user specified group identifier for easy querying
   - `skip-grouping` - Instead of enqueueing the tests as a logical group, they're effectively batched.
- Add a new `group:fetch` command that fetches a group from remote using an identifier. Options:
   -   `group-identifier` - the identifier for an existing group.
- Adds a new `group:show` command which fetches and shows the currently stored group if it exists.

### Grouping Local Tests
The only difference with how local tests(activation & e2e) tests are handled is that they run on the user's local device instead of on QIT's infra. When the `group:run` command is executed, the CLI will trigger the remote tests on QIT infra and sequentially run all the local tests without any additional input from the user. 

### Example Flow
**Add the run to a group**
`qit run:activation automatewoo --group --extension_set compatibility`

**Register the test group**
`qit group:register --group-identifier commit-1234567890`

**Run the test group**
`qit group:run`

**Response from the Manager**
```
Group enqueued on QIT servers!
Group ID: <id>
Group Identifier: commit-1234567890
--------------------------------
Test Run ID: <test_run_id>
Test Type: Activation
Test Results Manager URL: <url>
```

### Caveat

- If a group run has been triggered, any local run that matches the parameters of a group run WILL be picked up. To avoid this scenario, a user must pass the `--no_group` flag when starting their test run.

### Testing Instructions
1. Deploy the counterpart manager branch to staging
2. Checkout this branch and switch your CLI to staging
3. Add a remote and local test using the `--group` flag.
4. Add a duplicate for each and confirm that it doesn't get added.
5. Try a slight variation of the already added tests and confirm it gets added.
6. Run `qit group:show` and confirm that it outputs the added tests.
7. Run `qit group:register` and confirm that it gets registered and the the output includes the group id and identifier along with test run information
8. Try to add a new test and confirm that you aren't able to.
[...]

PR #275, opened 2025-03-26
Fix extension set resolution

singerb:
Pulls a small change from https://github.com/woocommerce/qit-cli/pull/261 out into a separate PR to fix extension set resolution.

Without this, you should see something like `Could not find Woo Extension with id 0.` when you run `./qit run:activation --extension_set=compatibility <extension>`; with the PR applied (and built) it should run properly.


zhongruige (approved):
Haven't tested but the changes look straight forward to me, thanks @singerb!

PR #276, opened 2025-03-27
Update from ubuntu-20.04 to ubuntu-latest

Luc45:
20.04 is going to be removed from GitHub, so we need to upgrade.


Luc45:
Merging this as there's virtually nothing to review.

PR #277, opened 2025-03-27
Also account for ts spec files

Luc45:
Playwright spec files can also be written in TypeScript.

### Testing Instructions

- `qit scaffold /tmp/foo`
- Paste this in `/tmp/foo/example.spec.ts` (See it ends in TS, not JS!)

```js
/**
 * my-plugin-menu.spec.ts
 *
 * Demonstrates various TypeScript features in a Playwright test:
 * - Enums
 * - Interfaces
 * - Typed function parameters
 * - Usage of generics (optional, shown in a helper function)
 */
import { test, expect, Page } from '@playwright/test';
import qit from '/qitHelpers';

/** An enum representing different WordPress roles (for example). */
enum UserRole {
  Admin = 'administrator',
  Editor = 'editor',
  Subscriber = 'subscriber',
}

/** 
 * Describes the shape of our test scenario. 
 * You can add more fields as needed for your test data. 
 */
interface PluginMenuTestCase {
  role: UserRole;
  menuTitle: string;
  expectedHeading: string;
}

/**
 * A strongly-typed helper function demonstrating how you might
 * use generics in TypeScript. We’ll just format a log statement here,
 * but generics can be used for any data type in real scenarios.
 */
function logTestCase<T>(item: T): void {
  // For demonstration, we’ll just console.log the item.
  // In a real test, you might do something more complex with it.
  console.log('Running test case:', item);
}

/**
 * A helper function demonstrating typed parameters.
 * Realistically, you might do additional logic here before calling `qit.loginAsAdmin`.
 */
async function loginAsRole(page: Page, role: UserRole): Promise<void> {
  // If we had different login strategies by role, we’d handle it here.
  // For demonstration, call the existing qit.loginAsAdmin.
  // (You could also have different calls like qit.loginAsEditor, etc.)
  if (role === UserRole.Admin) {
    await qit.loginAsAdmin(page);
  } else {
    // In a real scenario, you might implement logic for other roles.
    // We'll just do admin login for demonstration.
    await qit.loginAsAdmin(page);
  }
}

/**
 * An array of test cases to demonstrate TypeScript’s strong typing.
 * You can add more roles, menu titles, or expected headings here.
 */
const testCases: PluginMenuTestCase[] = [
  {
    role: UserRole.Admin,
    menuTitle: 'My Plugin Menu',
    expectedHeading: 'Dashboard',
  },
];

test.describe('My Plugin Menu E2E Tests', () => {
  for (const { role, menuTitle, expectedHeading } of testCases) {
    test(`Menu should be visible to role: ${role}`, async ({ page }: { page: Page }) => {
      logTestCase({ role, menuTitle, expectedHeading });
      
      // Log in as the specified role
      await loginAsRole(page, role);

      // Validate the expected heading is present
      await expect(
        page.getByRole('heading', { name: expectedHeading })
      ).toBeVisible();

      // (Optional) Check for the plugin menu. In a real test, remove comments:
      // await page.getByRole('link', { name: menuTitle, exact: true }).click();
      // await page.waitForLoadState('networkidle');
      // await expect(page.locator('h3')).toContainText('Welcome to My Plugin!');
    });
  }
});
```
- Run `qit run:e2e automatewoo /tmp/foo`
- See that it runs as expected, including the TS one.


zhongruige (approved):
Great catch! Thanks for getting this in.

PR #278, opened 2025-03-28
Normalize activation test self-test snapshots

Luc45:
Adds bits and pieces of the activation test output for snapshot testing used on our self-tests.

Some of these changes came in after we updated Playwright to 1.51.1


zhongruige (approved):
Nice, changes look good, thanks for this!

PR #288, opened 2025-04-17
Ignore notices that are not coming from the SUT on Activation Test type

Luc45:
This PR:

- Ignores the following deprecation notices that were being triggered by Query Monitor and would show up on other plugin's test reports on WordPress 6.8.0 (released 2 days ago)
  - `Hook setted_site_transient is deprecated since version 6.8.0! Use set_site_transient instead.`
  - `Hook setted_transient is deprecated since version 6.8.0! Use set_transient instead.`
- Similarly, skip logging noisy notices from `translation loading for the (.*) domain was triggered too early` when this is emitted by a plugin that is not the SUT

PS: There's some leftover noise in the snapshots of notices, but this is coming from the STDOUT of the Activation Test. In the "Rethinking Custom E2E Tests" PR, notices are no longer printed to STDOUT, so I won't tackle this here.


zhongruige (approved):
Changes look forward to me, thanks for tackling this @Luc45!

PR #290, opened 2025-04-18
Handle Plugin-Theme dependency

Luc45:
Plugin-Theme dependency relationship can be declared through WCCOM metadata.

This PR updates the CLI to take into theme dependency into account, if they are explicitly declared as WCCOM metadata.


zhongruige (approved):
Nice, this is a great update to get in. Thanks @Luc45!

PR #293, opened 2025-04-24
Self-test enh: allow to use --one-of-each to test that all test types are working

Luc45:
I've added this as a convenience to test the removal of Setup PHP, it just runs one test of each type for the self-test, which is just a convenience to run one scenario instead of multiple scenarios per test type.


zhongruige (approved):
Nice, this is a handy improvement. Thanks @Luc45!

PR #295, opened 2025-04-25
Fix logic that skips auto-activating theme is SUT is theme

Luc45:
Self-tests [revealed](https://github.com/woocommerce/qit-cli/actions/runs/14666806533/job/41163500379) that [this PR](https://github.com/woocommerce/qit-cli/pull/290) I merged yesterday introduced a regression in themes that depends on other themes. This has been fixed here.

**Testing Intructions:**

- Checkout this branch
- Point CLI to production
- Run `php QITSelfTests.php update activation parenttheme`
- See that it passes


zhongruige (approved):
LGTM! Thanks @Luc45!

PR #297, opened 2025-05-07
Using ubuntu-latest for workflow

Abdalsalaam:
Ubuntu 20.04 Actions runner image is deprecated, this will prevent running [QIT Self-Tests - Group Tests](https://github.com/woocommerce/qit-cli/actions/runs/14870632280) workflow

### DIFF
--- a/.github/workflows/qit-self-group-test.yml
+++ b/.github/workflows/qit-self-group-test.yml
@@ -13,7 +13,7 @@ on:
 
 jobs:
   self_test_custom_test:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-24.04
     env:
       NO_COLOR: 1
       QIT_DISABLE_ONBOARDING: yes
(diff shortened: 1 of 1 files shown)


.github/workflows/qit-self-group-test.yml
```diff
@@ -13,7 +13,7 @@ on:
 
 jobs:
   self_test_custom_test:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-latest
```
zhongruige: Thank you for the PR @Abdalsalaam! Can you please make one minor change? If we can use `ubuntu-24.04` instead of `latest` that would be preferable as this will match the logic on the QIT back end.

```suggestion
    runs-on: ubuntu-24.04
```

Thank you!

zhongruige (changes requested):

.github/workflows/qit-self-group-test.yml
Abdalsalaam: @zhongruige Thank you for reviewing this!

zhongruige (approved):
Thanks so much for getting in the changes @Abdalsalaam! LGTM!

PR #300, opened 2025-05-15
Added Group json option

MrJnrman:
This PR adds the `--json` option to the `group:fetch` command to render the output as raw json.

### Changes
- Add `--json` flag 
- Display the test group when clearing it
- If a completed test group exists when trying to enqueue a new test, it will be displayed then cleared in order to complete the action.

### Testing Instructions
1. Checkout this branch 
2. Enqueue 2 or more tests in a group
3. Run the group
4. Run `qit group:fetch -i <group-identifier>` and confirm you get nicely formatted output.
5. Run `qit group:fetch -i <group-identifier> --json` and confirm you get raw json.
6. Clear the group and confirm that you get an ouput.
7. Enqueue and run a new group and wait until its completed.
8. Try to enqueue a group again and confirm that it displays the old group and completes the action


Luc45 (approved):
Thanks!

src/src/TestGroup.php
```diff
@@ -535,4 +540,35 @@ public function fetch( string $group_identifier ): array {
 
 		return json_decode( $response, true );
 	}
+
+	public function output_group( OutputInterface $output ): void {
+		$group = $this->get();
+		if ( ! array_key_exists( 'group_identifier', $group ) ) {
```
alopezari, line 546: Should we ensure that $group is an array here and in other similar checks in order to make it more robust? Something like:
```
if ( is_array( $group ) && ! array_key_exists( 'group_identifier', $group ) )
```

alopezari:
Thanks for creating this PR @MrJnrman!

Please don't consider my questions/comments as blockers and go ahead if you're happy with Lucas' approval!

Just wanted to test this PR as an opportunity to learn more about groups. Testing instruction step 2 says:
```
Enqueue 2 or more tests in a group
```

How can you do that?

src/src/TestGroup.php
MrJnrman: We don't need to. The function `get` method on the TestGroup class returns an empty array if the group doesn't exist and everywhere else where we fetch the group directly from the cache has defensive checks in place

MrJnrman:
@alopezari I'm happy to share. To add a test to a group all you have to do is append `--group` to a command used to run a test. What that will do is differ execution and add the test to a locally declared group. Once you've populated your group, you can execute all of them at once by running `qit group:run`. If there are any e2e/activation tests in your group, they'll be executed automatically and the QIT manager will be updated once they're complete. 

You can also check the docs for more info: https://qit.woo.com/docs/managed-tests/group-tests

PR #303, opened 2025-06-06
QITCommand Abstract class

Luc45:
This is the second PR to break down the monolith PR. It introduces the `QITCommand` abstract class, from which all Command classes extends.


zhongruige:
Any specific testing I should focus on for this one @Luc45? Or do you think just going through and checking for any regressions on current behavior (auth, running tests etc) should be sufficient?

src/src/Commands/ConnectCommand.php
```diff
@@ -45,7 +44,7 @@ protected function execute( InputInterface $input, OutputInterface $output ): in
 
 		$command = $this->getApplication()->find( AddPartner::getDefaultName() );
 
-		return $command->run( new ArrayInput( [] ), $output );
```
alopezari: In other methods we're returning the same SUCCESS or FAILURE constant.

Here, it seems we're returning a different thing than before (before we used to run a command and now we return SUCCESS. Is this expected?

src/src/Commands/OpenCommand.php
```diff
@@ -2,27 +2,27 @@
...
 			->setDescription( 'Open a test run result in the browser.' )
 			->addArgument( 'test_run_id', InputArgument::REQUIRED );
 	}
 
-	protected function execute( InputInterface $input, OutputInterface $output ): int {
+	protected function doExecute( InputInterface $input, OutputInterface $output ): int {
 		$command = $this->getApplication()->find( GetCommand::getDefaultName() );
-
-		return $command->run( new ArrayInput( [
+		$command->run( new ArrayInput( [
```
alopezari: Same as above, we used to run a command and now we return SUCCESS. Is this expected? Also, what if the command fails? It would stop the execution and we wouldn't return SUCCESS here, right?

alopezari (commented):
Great job improving the static analysis and building a more robust architecture in this project @Luc45!

I asked a couple of questions just in case, but it looks good to me.

Regarding testing, I second @zhongruige question! Please let us know when it is a proper time to test these changes.

src/src/Commands/OpenCommand.php
Luc45: Thanks for the thorough review, that was indeed a typo. Fixed.

src/src/Commands/ConnectCommand.php
Luc45: Thanks for the thorough review, that was indeed a typo. Fixed.

src/src/Commands/OpenCommand.php
Luc45: I've also double-checked the rest of the PR, didn't see a leftover typo like those.

Luc45:
> Any specific testing I should focus on for this one @Luc45? Or do you think just going through and checking for any regressions on current behavior (auth, running tests etc) should be sufficient?

This is just breaking up the monolith to reduce "files changed", so no need to take this PR too seriously for now

Luc45:
This will be merged in a long-lived branch for QIT CLI 1.0 - not trunk

zhongruige:
> This will be merged in a long-lived branch for QIT CLI 1.0 - not trunk

Ah okay cool, since it's going there I am less worried and the changes seem isolated enough, too. So, I'm happy to add my approval to this!

zhongruige (approved):

alopezari (approved):
Cool! In that case, let me approve it too 🙌

PR #310, opened 2025-06-19
Move tests to "integration" directory

Luc45:
- This PR moves `_tests/custom_tests` to `src/tests/integration`.
- Moves `./src/tests` to `./src/tests/unit`
- And tweaks the `json-filter.php` so that it don't omit errors if `--json` is passed

This PR is part of breaking down the monolith QIT CLI 1.0 PR.


Luc45:
It's fine if the PHPUnit tests fails here, this is WIP and is just for breaking down the monolith.

MrJnrman (approved):
Makes sense. Thanks for the changes!

PR #311, opened 2025-06-24
Sync WC 10.0 tests

zhongruige:
Double check snapshots look good.


Luc45 (approved):
Snapshots looks good :+1:

PR #296, opened 2025-05-02
Explicitly run plugin pre_activation tests in the context of Activation Tests

MrJnrman:
This PR ensures that we explicitly run the pre activation step for all plugins in the activation stack in the context of an Activation test. This is to ensure that the default bootstrap behavior doesn't occur too early in the setup phase.

### Testing Instructions
1. Checkout this branch 
2. Connect your cli to prod
3. Run an activation test with 2 or more additional plugins that have custom tests set up. You can use this command `./qit run:activation woocommerce-gateway-stripe --wp 6.8.1 --woo 9.8.3 --php_version 7.4 --pw_test_tag host-plan -p 'woocommerce-deposits' -p 'woocommerce-payments' -p 'woocommerce-paypal-payments' -v`
4. Confirm the test executes successfully.

PR #316, opened 2025-07-10
Set `pre_activation` test tag additional plugins for e2e tests that run the woocommerce activation tag

MrJnrman:
Follow up to https://github.com/woocommerce/qit-cli/pull/296

This PR ensures that we explicitly run the pre activation step for all plugins in the activation stack in the context of an e2e test that used to run a woo activation test. This is to ensure that the default bootstrap behavior doesn't occur too early in the setup phase.

### Testing Instructions
1. Checkout this branch 
2. Connect your cli to prod
3. Run an activation test with 2 or more additional plugins that have custom tests set up. You can use this command `./qit run:e2e woocommerce activation --wp=6.8.1 --woo=9.9.5 --php_version=8.4 --plugin=woocommerce-gateway-stripe `
4. Confirm the test executes successfully.


zhongruige (approved):
Nice, tested great for me, thanks @MrJnrman!

Luc45 (approved):

PR #319, opened 2025-07-16
Snapshot changes with updated packages

Luc45:
This PR is the snapshot changes with the updated docker images with WP CLI and Playwright updates, and the AI feature disabled in a feature flag.

It is looking good, which some small changes that are not related to any of these PRs.


zhongruige (approved):
Nice, looks good to me!

PR #320, opened 2025-07-21
Add remote handling for performance tests

alopezari:
### Description
This PR enhances the local performance testing functionality by adding support for remote tests handling. The main implementation includes improved test download logic, better environment management, and refined command arguments alignment with the existing E2E test commands.

This PR extends and improves the original work that was introduced formerly by [this other PR](https://github.com/woocommerce/qit-cli/pull/315).

### Changes added in this PR
- **Remote tests handling:** Added logic to properly handle remote tests during local performance test runs, including improved test downloading and environment setup.
- **Command alignment:** Aligned run:performance command arguments with run:e2e for consistency across test types.
- **Environment termination:** Added proper cleanup logic for performance environments when tests finish or are aborted.
- **Docker build fix:** Fixed repository configuration for older Debian versions (Buster) to resolve build issues with PHP 7.2

### Testing instructions
- Checkout to this branch and make sure your QIT-CLI is pointing to a Compatibility Dashboard version that contains the changes from [this other PR](https://github.com/Automattic/compatibility-dashboard/pull/1408) (either locally or on staging).
- Upload the performance tests with `php src/qit-cli.php tag:upload woocommerce:{tag name} {performance tests path}`. For example, `php src/qit-cli.php tag:upload woocommerce:default ../compatibility-dashboard/ci/tests/performance/custom-test/ performance`.
- Make sure the performance tests are listed when checking the tags with `php src/qit-cli.php tag:list`. You should see something like:
```
+-----------------------------+---------------------------------------------------+-------------------+--------+
| Slug                        | E2E Tests                                         | Performance Tests | Type   |
+-----------------------------+---------------------------------------------------+-------------------+--------+
| woocommerce                 | nightly, activation                               | default           | plugin |
+-----------------------------+---------------------------------------------------+-------------------+--------+
```
- Run the performance tests with `php src/qit-cli.php run:performance woocommerce --verbose`
- Ensure the performance environment that was created for running the tests no longer exists, and it was successfully cleared with the previous command.
- Check that all the test artifacts were created successfully and you can open them.
- Run the activation tests with `php src/qit-cli.php run:activation woocommerce --verbose`, in order to catch potential regression issues.
- Please consider testing other flows that might have not been defined here.

PR #323, opened 2025-07-28
Add new CLI argument to select which k6 test to run

alopezari:
This PR adds a new optional argument to the `run:performance` command named `--k6_test_file`  to select which k6 test to run, in case the performance tests include multiple tests or follow a specific architecture involving multiple files.

Additionally, it updates the logic that identifies k6 tests to be less restrictive as it was only considering files with the `.k6.js` extension as k6 tests, but these can be stored in simple `.js` files.

### Changes added in this PR
- Removed former docker user args defined K6DockerConfig as it was causing conflicts with k6 browser.
- Updated RunPerformanceTestCommand to define a new `--k6_test_file` argument within the `run:performance` command that will point to what exact k6 test file to run.
- Updated PerformanceEnvInfo and K6Runner to use this new `k6_test_file` argument.
- The k6 Dockerfile was removed, as it was part of the initial implementation that handled k6 tests locally, instead of storing them in the Compatibility Dashboad. As this was completed in #320, the Dockerfile is no longer needed.

### Testing Instructions
This PR should be tested together with https://github.com/Automattic/compatibility-dashboard/pull/1417.

- Make sure Compatibility Dashboard is running locally with the changes from [this branch](https://github.com/Automattic/compatibility-dashboard/pull/1417).
- Ensure QIT-CLI uses the latest changes from this branch.
- Upload the performance tests with `php src/qit-cli.php tag:upload automatewoo:{tag_name} {performance_tests_path} performance`. For example, `php src/qit-cli.php tag:upload automatewoo:demo ../compatibility-dashboard/ci/tests/performance/custom-test/ performance`.
- Run the performance tests with `php src/qit-cli.php run:performance automatewoo demo --k6_test_file=frontend.js --verbose`.
- Ensure the tests pass.
- Please test around this PR.


zhongruige:
For the tests at `{performance_tests_path}`, is there an example of what these should look like as far as folder and file structure? Could be useful to include as part of the test instructions (also something to make sure we update [our docs](https://qit.woo.com/docs) with too).

alopezari:
Good point @zhongruige!

There isn't any at the moment as we keep iterating on the implementation as we go through it and it might change. I've created a new task to cover it once we're close to project completion:
[QIT-758: Add performance tests documentation and examples](https://linear.app/a8c/issue/QIT-758/add-performance-tests-documentation-and-examples)

zhongruige:
Ah awesome okay that was the only hold up I saw while testing this, so thanks for confirming!

src/src/LocalTests/Performance/Runner/K6DockerConfig.php
```diff
@@ -45,6 +44,7 @@ private function get_base_docker_args( PerformanceEnvInfo $env_info, string $con
 			'--rm',
 			'--init',
 			'--add-host=host.docker.internal:host-gateway',
+			'--cap-add=SYS_ADMIN',
```
zhongruige: Is there a reason we need to have the SYS_ADMIN capability here? Concerned since this provides root access to the host, I'm not sure if we need or want to do this.

alopezari: Good catch! I had trouble trying to run k6 browser within the container and tried adding different container options based on blogs and AI recommendations (this was one of them) until it worked. I will double check if we do need it.

alopezari: Hi @zhongruige!

I verified that the SYS_ADMIN capability didn't actually solve the issues I was finding originally when setting up the environment for k6 browser, the issue was caused by trying to run k6 browser in headed mode through this setup (it's not possible as k6 runs inside a docker container). There is a workaround but it requires installing k6 locally, just for dev purposes.

In any case, we can safely remove SYS_ADMIN capability, which we did in [23dd054](https://github.com/woocommerce/qit-cli/pull/323/commits/23dd05402bf8d6a486d4e1569afbb473891f81ab).

zhongruige: Sounds great, thanks for the context and looking into this @alopezari! Just to double check, are there any other changes here coming in or is this all set to test/review?

alopezari: Yes, it's ready for testing/review! Any further changes will land in upcoming PRs.

zhongruige:
@alopezari I've been seeing this crop up again after testing around this PR, want to check if you're seeing the same or if it's something in my workflow--I used the binary to test this PR but now I'm using the one at my path, not sure if it makes a difference though:

```
 qit switch staging


 [ERROR] Invalid environment type.
```

```
qit run:activation automatewoo


 [ERROR] Invalid environment type.
```

alopezari:
Hmm interesting @zhongruige! I haven't seen this behaviour since we made the QIT more robust by falling back to e2e environment if an unknown environment was set in the QIT cache:
https://github.com/woocommerce/qit-cli/pull/315/commits/434f472bc2278ca46ca6bb8e9eb7db0895efc045

Is it reproducible? Could you please share the steps to reproduce?

zhongruige:
Wasn't able to replicate it and tried a few different things, so I think we're good here, happy to add my approval @alopezari !

zhongruige (approved):
LGTM!

PR #326, opened 2025-08-05
Add more tests for vendor code

Luc45:


zhongruige (approved):
Thanks for getting in these self tests!

PR #327, opened 2025-08-05
Snapshot tweaks

Luc45:
Snapshot tweaks for our self-tests


Luc45:
Merging without review as it's just snapshots

PR #328, opened 2025-08-07
Include Basic Auth plugin in the performance environment setup

alopezari:
As part of performance tests, we need to send requests to the SUT's REST API. In order to make it simple, the performance environment setup logic has been updated to include the install and activation of the Basic Auth plugin, which streamlines the REST API setup of a WordPress site.

Additionally, the k6 image used in our setup has been switched from `latest` to `master`, which is the stable version.

### Testing instructions
- Checkout to this branch.
- Set up a performance environment with `php src/qit-cli.php env:up --environment_type=performance`.
- Go to the plugins page (`/wp-admin/plugins.php`) and verify that the **JSON Basic Authentication** plugin is installed and activated.


zhongruige (approved):
Changes look good to me and are straightforward, thanks @alopezari !

PR #329, opened 2025-08-11
Enable cash-on-delivery payment method by default in performance environment

alopezari:
In order to run the [performance tests for the checkout flow](https://github.com/Automattic/compatibility-dashboard/pull/1438), we need to ensure at least one payment method is enabled.

This PR enables the `cash-on-delivery` payment method during the performance environment setup through WP-CLI.

### Testing instructions
- Checkout to this branch.
- Set up a performance environment with `php src/qit-cli.php env:up --environment_type=performance`.
- Go to the offline payment methods page (`/wp-admin/admin.php?page=wc-settings&tab=checkout&section=offline`) and verify that `cash-on-delivery` is enabled.


zhongruige (approved):
Nice looks good and I like the use of WP CLI here!

PR #334, opened 2025-08-28
Add base data in performance tests environments

alopezari:
In order to run performance tests properly, we need to load a bunch of data in the test environment, since performance bottlenecks frequently appear when the system is under high load. According to recent internal researches, it seems 1000 products and 100000 orders would be a good baseline to start testing the performance of WooCommerce sites.

This was a challenge because creating such amount of data was very slow with the methods we had available (e.g. WooCommerce Smooth Generator).

This PR follows a different approach to generate that amount of data in a short time, creating the environment with all the data in approximately 🟢 **1 minute** 🟢 .

### The solution
Instead of using WooCommerce Smooth Generator or the API REST, we decided to create the data directly in the database. To do so, we:
1. Created a plugin that captured and logged the db queries done while performing user flows.
2. Performed the simple product, variable product and order creation flows manually on a new site in order to capture the db queries with the plugin created previously.
3. We created another plugin that used the info from step 2 to create orders and products sending the same db queries we captured initially. This new plugin works similar to WC Smooth Generator but it's way faster.
4. We created a new site with 1000 variable products with variations and 100000 orders, and we created a SQL dump from it.
5. We uploaded the SQL dump to the prod Compatibility Dashboard site, in order to make it available to all QIT-CLI environments. **Note: We will handle the automatic generation and upload of SQL dump files in a separate PR.**
**6. We created this PR, which downloads the SQL dump file and imports it in the database.**

### Changes in this PR
* The PerformanceEnvironment class was updated to download a SQL dump with 1k products and 100k orders and import it in the environment database every time a new performance environment is created.

### Testing instructions
* Make sure you're using a Compatibility Dashboard version that contains the changes from [this PR](https://github.com/Automattic/compatibility-dashboard/pull/1460).
* Checkout QIT-CLI to this branch and point it to the Compatibility Dashboard version from the previous step, either local or staging, as long as it contains the changes from the PR above.
* Set up a performance environment with `php src/qit-cli.php env:up --environment_type=performance --woo=stable`.
* Check that no errors were logged in the terminal.
* Once the site is created, go to the admin panel, open the products page and check that there are 1000 products. Then open the orders page and verify that there are 100000 orders.


alopezari:
PHPUnit tests are failing, but they also fail on `trunk`, so I assume it's not related to this PR:
https://github.com/woocommerce/qit-cli/actions/runs/17294286139/job/49088623078#step:6:88

src/src/LocalTests/Performance/Environment/PerformanceEnvironment.php
```diff
@@ -345,4 +348,37 @@ protected function additional_output(): void {
...
+	 * Download and import the performance database dump.
+	 */
+	private function generate_base_data(): void {
+		$this->output->writeln( '<info>Generating test products and orders...</info>' );
+
+		// Download and import in one command to avoid storing large files.
+		$import_command = implode( ' && ', [
+			'cd /tmp',
+			'echo "Downloading performance database dump..."',
+			'curl -L -o woocommerce_dump.sql.zip https://qit.woo.com/wp-content/uploads/qit-env-db-dumps/woocommerce_dump.sql.zip',
```
zhongruige: Since this is technically a public repo, what do you think about putting this in a secret? It isn't necessarily exposing anything, but just a thought!

alopezari:
Hi @zhongruige , @MrJnrman !

I iterated on the PR review suggestion to improve the way to download the SQL dump file. You can check it here:
[6570b1b](https://github.com/woocommerce/qit-cli/pull/334/commits/6570b1bd5961b57d5eff94874d60fe76ae96e7f0)

MrJnrman (approved):
Thanks for the changes @alopezari!

PR #336, opened 2025-08-29
Build performance reporting system

alopezari:
Prepare the reporting system for performance tests that consists of:
- A summary JSON report.
- An extended JSON report with extensive data.
- An HTML report for visual checking.
- A proper db report in `cd_test_result_json` column from the `wp_qit_test_runs` table.
- A new `cd_performance_results` column in the `wp_qit_test_runs` table that contains summarized data needed for quality tasks, as well as for calculating the performance score later in this project.

### Changes added in this PR
- Added proper handle and built-in HTML report generation in `K6DockerConfig`.
- Removed former dummy HTML report in `PerformanceTestResult`.
- Added proper naming of existing reports (`result.json`, `result-extended.json` and `dashboard-report.html`.
- Updated `LocalTestRunNotifier` to exclude performance tests from the Playwright to Puppeteer test result conversion and generate a summary that gets stored in the new `cd_performance_results` column.

### Testing instructions
This PR shares the same testing instructions than https://github.com/Automattic/compatibility-dashboard/pull/1468 (both need to be tested together).


zhongruige (approved):
Hi @alopezari! Just a friendly reminder to build the binary with these changes. Other than that, looks good to me!

alopezari:
Thank you @zhongruige! I tend to forget about it 🤦 will investigate how to set a check that automatically flags when you need to create a new binary 🙌

alopezari:
PS: I'm also investigating the failing unit test, I don't think it's related to this PR as I've seen it in other places, but it's a bit annoying to see it failing always and we need a test suite we can trust.

zhongruige:
Sounds good! Feel free to punt on the unit test for another PR if you want, we can try looking at it separately

PR #338, opened 2025-09-02
Fix PHPUnit issue due to differences in ZipValidator behaviour depending on the PHP version

alopezari:
This PR fixes a recent PHPUnit error that made our tests fail in the pipelines.

Example:
https://github.com/woocommerce/qit-cli/actions/runs/17402337049/job/49398116176#step:6:88

If we investigate the recent executions of this test in our GitHub Actions, we'll find out it works for PHP 7.2 but fails for PHP 8.3/8.4.

Therefore, this error seems to be caused by the PHP version where it runs, and we need it to be compatible with both, so we have 2 options here:
- Check that there is an `Exception` only, or...
- Check the 2 exception types we know so far, and keep extending it as long as we find new ones with new PHP changes. This involves some maintenance but will slightly increase the test coverage. 

In this PR, we opted for the 2nd approach, as you can see in [Files changed](https://github.com/woocommerce/qit-cli/pull/338/files), but feel free to suggest the first one if you consider it more appropriate.

### Testing instructions
No manual testing is needed, just ensure the PHPUnit tests passed.


zhongruige (approved):
Nice, clean and simple way to fix the test, love it!

PR #339, opened 2025-09-03
Add new baseline logic in performance tests

alopezari:
## Description

This PR adds a new flow to the performance tests. Now, every time we run the `run:performance` command, it will run performance tests against a site with WooCommerce only, then it will run the tests again against a site with WooCommerce and the SUT, and it will store both results in the `cd_performance_results` from the `wp_qit_test_runs` table. It is also possible to run the performance test without the baseline with the `--no_baseline` argument.

The Compatibility Dashboard will use this data to show a proper test report exposing the performance metrics for both test runs as well as a comparison between baseline vs SUT metrics to see how the SUT impacted the performance of WooCommerce.

Note: Right now the metrics fluctuate too much, this is because we have to find a way to provide stable metric values, which I plan to handle in a separate task.

### 1. **New Classes Created**
- **MetricsExtractor.php** - New simplified metrics extraction class with a single `extract_metrics()` method
- **PerformanceComparisonService.php** - New service to handle performance metric comparisons between baseline and extension tests
- **PerformanceTestConfig.php** - New configuration class to centralize performance testing constants and metrics definitions

### 2. **Enhanced PerformanceTestResult**
- Added baseline tracking functionality with `is_baseline` property
- Added `set_baseline()` and `is_baseline()` methods
- Added `set_baseline_result()` and `get_baseline_result()` methods to link baseline results to main test results
- Enhanced `process_k6_results()` to skip raw metric definitions
- Modified `calculate_summary_statistics()` to process all metrics including Web Vitals
- Added `write_result_json()` method for backwards compatibility

### 3. **Refactored PerformanceTestManager**
- Added support for running baseline tests (WooCommerce only) before extension tests
- Added `run_baseline_tests()` method to execute baseline performance tests
- Added `create_baseline_environment()` method to create test environment without the extension
- Added `combine_results()` method to merge baseline and extension results with comparisons
- Integrated the new PerformanceComparisonService for metric comparisons

### 4. **Updated LocalTestRunNotifier**
- Modified `extract_combined_performance_metrics()` to handle baseline results
- Updated performance results structure to include baseline, extension, and comparison data
- Simplified metric extraction to use the new MetricsExtractor class

### 5. **Minor Updates**
- Added `run_baseline` flag to PerformanceEnvInfo
- Updated RunPerformanceTestCommand to handle the `--no-baseline` option
- Modified the performance results JSON structure sent to the compatibility dashboard

## Testing instructions

You can reuse the testing instructions from [this Compatibility Dashboard's PR](https://github.com/Automattic/compatibility-dashboard/pull/1469), as they need to be tested together.


zhongruige (approved):
Looks good, just a reminder to build the binary before merging @alopezari

PR #340, opened 2025-09-05
Update self-tests snapshots.

alopezari:
After the recent changes, self-tests started failing due to outdated snapshots.

This PR aims to update them to fix the these tests.


zhongruige (approved):
LGTM thanks for updating!

Luc45:
Thanks for updating, but updating is not enough, as those values are dynamic and change on every run. We must normalize them.

alopezari:
Thanks for the info @Luc45! It was the first time I worked on this, so I sticked to [this documentation](https://github.com/woocommerce/qit-cli/blob/trunk/\_tests/managed_tests/README.md#update-snapshots). What are the steps for normalizing them?

PR #341, opened 2025-09-05
Normalize and update snapshots

Luc45:
Follow-up to https://github.com/woocommerce/qit-cli/pull/340, it normalizes the dynamic fields to make sure snapshots are consistent  and update them.


Luc45:
I'll be merging this right away as it's just snapshot changes.

Luc45:
The Custom Tests self-tests is expected not to pass, it will be removed once the Test Packages work goes in.

PR #342, opened 2025-09-06
Add new argument for switching between running performance tests on CI or locally

alopezari:
This PR add the necessary changes to QIT-CLI to run performance tests on CI, in conjunction with https://github.com/Automattic/compatibility-dashboard/pull/1489.

### Changes added in this PR
- `RunPerformanceTestCommand` was updated to support both CI (by default) and local (by using the option `--local`) execution.
- `PerformanceTestResult` was updated to fix an issue with results folder creation when overriding the results path with `QIT_RESULTS_DIR`.
- `K6DockerConfig` was updated to provide better permission handling on the container that runs k6 tests, as it was causing some issues when writing the results files.

### Testing instructions
Please follow the testing instructions described in the [corresponding CD PR](https://github.com/Automattic/compatibility-dashboard/pull/1489), as both need to be tested together.


MrJnrman (approved):
Looks great and tested well against the manager PR

PR #343, opened 2025-09-08
Remove the performance metrics comparison from QIT-CLI as it will be done entirely in the Compatibility Dashboard

alopezari:
Before this PR, the QIT-CLI retrieved the performance metrics from performance test results, processed the metrics and created a comparison object for the Compatibility Dashboard to consume.

While working on the Performance Score calculation logic, I've decided to leave the QIT-CLI as a metrics provider and do the processing and comparison entirely within the Compatibility Dashboard as an architecture improvement.

### Testing instructions
No extra testing is needed as it will be covered in the corresponding CD PR - https://github.com/Automattic/compatibility-dashboard/pull/1474.


zhongruige (approved):
Changes tested great for me, thanks @alopezari!

PR #344, opened 2025-09-10
Allow performance tests to run on multiple iterations and get the average of metrics for better stability

alopezari:
Performance tests retrieves performance browser metrics from a set of user flows. These metrics fluctuate depending on network, hardware, etc.

One of the tweaks we did in the process is ensuring the baseline is run in the same thread than the SUT test, so that they share same cores and even hardware, in case of CI Actions.

In order to help mitigating that metrics fluctuation, this PR aims to run both the baseline and SUT performance tests a given number of iterations and calculates the average value of the metrics, in order to normalize them and make them more stable.

By default, 3 iterations will be run for each baseline and SUT test, but you can tweak this number by using the `--iterations` option in the `run:performance` command (for example, `--iterations=6`).

### Changes added in this PR
- Updated [RunPerformanceTestCommand.php](https://github.com/woocommerce/qit-cli/compare/add/performance-tests-multiple-iterations?expand=1#diff-a4dde50dee23600fa2fb30cd083a5f591b3872087414d2e9c69b266db5296579) to allow a new `--iterations` argument.
- Added a new [MetricAverager.php](https://github.com/woocommerce/qit-cli/compare/add/performance-tests-multiple-iterations?expand=1#diff-f725b5043c8937038a12816a91e3f4394a810d3dae2f51ddfb59221521a8cbe1) class in charge of calculating the average of multiple iterations.
- Updated [PerformanceTestManager.php](https://github.com/woocommerce/qit-cli/compare/add/performance-tests-multiple-iterations?expand=1#diff-443fcdea91ba6540319e1bf83e153b430f6ab91d2a653dd4283d16334063f3c3) to manage and handle the new iterations and average calculation process.
- Added new PHPUnit tests in [MetricAveragerTest.php](https://github.com/woocommerce/qit-cli/compare/add/performance-tests-multiple-iterations?expand=1#diff-35fa03b4628c7c768037eb0423d28be7775cc4c51d7ece3904363702c593640d) to cover the calculation of the average values of the performance metrics used for calculating the performance score.

### Testing instructions
- Load the changes from this branch and set up CD either locally or in staging.
- Run performance tests with default iterations with `php src/qit-cli.php tag:upload automatewoo:demo ../compatibility-dashboard/ci/tests/performance/ performance && php src/qit-cli.php run:performance automatewoo demo --k6_test_file=frontend.js --verbose`.
- Ensure the baseline tests ran 3 times and the extension tests ran 3 times.
- Ensure both the baseline tests and extension tests ran in different environments.
- Open the test report page in CD (example url from my local env: http://qit.test:8081/wp-admin/post.php?post=4832275&action=edit).
- Open also the test results folder that you'll see in the terminal logs at the end of the run:performance command execution (example: `/var/folders/j8/1djrt0mj0p56ph_dlrxghbz00000gn/T/qit-results-68c19466dec92`).
- Check that the **median** metric values of the `result.json` file match those from the CD test report page.
- Additionally, check that the **median** metric values you can see in the `result.json` file are the average value of the same median values for those metrics in the n iterations.
- Repeat testing with different combinations of iteration values.
- If you want, you can also check running multiple iterations without a baseline, by adding also the option `--no_baseline`. In that case, it should keep the number of iterations but running only the extension tests, i.e. without baseline.


zhongruige (approved):
This is a smart implementation, nice work @alopezari!

PR #347, opened 2025-09-22
Snapshot updates for 10.2

zhongruige:


zhongruige:
@alopezari just want to make sure this is on your radar, seems iterations might be showing for other test types (like API), might be worth looking into as a follow up:

https://github.com/woocommerce/qit-cli/pull/347/files#diff-de2f547654b7e7f914a217fe2ca27e1dd0ae159e56abce00dbd0667bcbd11dd4

Screenshot 2025-09-22 at 2 19 50 PM

alopezari:
Thanks for flagging this @zhongruige!

I've been investigating this and it seems a limitation rather than an issue added by the recent performance tests PRs.

As you can see, there are also other attributes related to other test types, like `malware_whitelist_paths`, `phpstan_level` or `test_variation`. Those belong to malware, security (I think so) and activation tests, respectively.

Right now our `[jsonSerializer](https://github.com/Automattic/compatibility-dashboard/blob/a6bbf4289676d9df828ec7adf69cfb52a6046a98/plugins/cd-manager/src/TestRuns/TestRun.php#L665)` function from `TestRun` gets all these attributes that are part of the TestRun object, no matter the test type, and this is why we can see it in those snapshots from the woo-api tests with empty or default values.

This might require a refactor of the existing implementation and it doesn't only impact the performance tests, so I think it's worth addressing it as a separate task (not urgent in my opinion)

zhongruige:
Agree with the assessment, we can tackle it later but since it isn't impacting any existing behavior we can tackle it later. Thanks @alopezari !

PR #350, opened 2025-09-23
Add default k6 test file and fix inconsistencies in run:performance arguments

alopezari:
This PR ensures the `default` test tag and `default.js` k6 test file are run by default if they're not defined as options or arguments when running performance tests.

Additionally, it resolves an inconsistency when setting the WordPress and WooCommerce versions when running performance tests, as our CI tests expect these arguments like `--wordpress_version` and `--woocommerce_version` but our tests that run locally define `--wp` and `--woo` instead, and these arguments are defined by parent classes that impact multiple test types, so I needed to find a workaround that solves the problem at this point until a proper refactor is done and has the least impact possible.

### Changes added in this PR
- Set `default` as the default test tag to be used when running performance tests if no test option is provided in the command.
- Set `default.js` as the default k6 test file to be used when running performance tests if no `--k6_test_file` argument is provided in the command.
- Set `--woocommerce_version` and `--wordpress_version` as the correct arguments to define the WP and WC versions in the test run, mapping them internally to `--wc` and `--wp` for local execution.
- Improved error logs for invalid test tags and k6 test files values.
- Updated snapshots.

### Testing instructions
- Set up the local environment.
- Make sure to use the QIT-CLI version from this branch.
- Upload the default test tag for any extension, like AutomateWoo:
```
php src/qit-cli.php tag:upload automatewoo:default ../compatibility-dashboard/ci/tests/performance/ performance
```
- Run the performance tests with default test tag for that extension, and ensure you define other WC and WP versions than default:
```
php src/qit-cli.php run:performance automatewoo --verbose --no_baseline --iterations=1 --wordpress_version=6.7.2 --woocommerce_version=10.1.1
```
- Ensure the local environment installs these WP and WC versions, and the tests run successfully because they can find the test tag and k6 test file even if they are not specified.
- Please test around this issue (consider running different combinations of existing / non existing test tags / k6 test files).


zhongruige (approved):
Looks good, seems it picked up the versions as expected:

Screenshot 2025-09-25 at 2 03 02 PM

Everything passed and the results look great too: https://stagingcompatibilitydashboard.wpcomstaging.com/?qit_results=1412409.Qv6BvqjobHNR7DxEGnHI9txOD0qJyexmLxxrp0gbu2ZolzNK6fy6fTn6UagjPpKS

Happy to add my approval!

PR #362, opened 2025-10-22
Fix download package duplicated log

alopezari:
There is a duplicated log in the step that downloads test packages which we can see, for example, when running activation tests. Example:

```
Running activation test scenario.
Fetching package metadata...
Downloading package: woocommerce/activation:latest
Downloading package: woocommerce/activation:latest
Installing npm dependencies...
```

That doesn't mean the `woocommerce/activation:latest` package is downloaded twice, it was just being logged twice in the terminal.

This PR fixes that and removes the duplicated log:

```
Running activation test scenario.
Fetching package metadata...
Downloading package: woocommerce/activation:latest
Installing npm dependencies...
```

### Testing instructions

- Load the changes from this branch.
- Ensure there is at least one package published for activation tests.
- Run activation tests for that package with `php src/qit-cli.php run:activation woocommerce`.
- Check that the logs only contain the line `Downloading package: woocommerce/activation:latest` once.


MrJnrman (approved):
Thanks for the quick fix! @alopezari

PR #364, opened 2025-10-24
Refactor SUT handling in performance tests

alopezari:
This PR refactors how the SUT variable is handled by performance tests in order to fix an issue when running on CI and make the implementation consistent with e2e tests.

Before this change, you could see the following error when running performance tests on CI:
- https://github.com/Automattic/staging-compatibility-dashboard/actions/runs/18777188767/job/53574448739#step:26:1155

After this change, the error is gone:
- https://github.com/Automattic/staging-compatibility-dashboard/actions/runs/18783490575/job/53595597386#step:26:1357

### Testing instructions
I don't think we need extra testing as we got testing deliverables on staging before and after these changes that we can just review.


zhongruige (approved):
Looks good to me, thanks @alopezari!

PR #366, opened 2025-10-24
Fix PHP version display

zhongruige:
This PR fixes the way we display PHP versions for PHP Compatibility tests--previously we didn't show the range and only the "base" version (which doesn't make much sense for this test). The new version is below:

Screenshot 2025-10-24 at 4 56 46 PM
Screenshot 2025-10-24 at 4 57 22 PM

---

FWIW I also played with a split out version, but I like the single line a bit better, though happy to switch it back depending on what folks think!

Screenshot 2025-10-24 at 4 52 20 PM
Screenshot 2025-10-24 at 4 53 12 PM

(note that this was before I removed the PHP version line, I would keep it out also for this version!)


MrJnrman (approved):
Thanks for these changes @zhongruige!

PR #369, opened 2025-10-27
Remove ": mixed" for PHP 7.4 compatibility

Luc45:
Fix typo, ":mixed" is only available in PHP 8.0


Luc45:
Merging this as it's super straightforward.

PR #373, opened 2025-11-07
Use the same performance environment for running both baseline and extension tests in order to reduce variance

alopezari:
Replaced the previous approach to run baseline vs SUT tests.

Before:
1. Baseline environment was created.
2. Baseline tests were run.
3. SUT environment was created.
4. SUT tests were run.

Now:
1. Environment is created
2. Baseline tests are run.
3. DB is restored in the environment by loading the db dump again.
4. SUT tests are run.

This process aligns SUT tests even more to baseline tests since everything would be running in the same container and under the same conditions.

Additionally:
- The code sets alias for WP/WC options was restored, since it was missing somehow.
- The latest version of the k6 docker image caused an issue where the metrics CLS and INP were not retrieved, so this PR fixes the image to version v1.3.0 (the previous one, which works properly).

### Testing instructions
1. Load the changes from this brach.
2. Publish the changes to the test package with `php src/qit-cli.php package:publish ../compatibility-dashboard/ci/tests/performance/test-package/ latest`
3. Run the performance tests with `php src/qit-cli.php run:performance automatewoo -vvv --iterations=1 --local`
4. Note that now only a single environment is created for both the baseline and SUT tests.
5. Ensure there are no regressions.


zhongruige:
This is a smart approach, I really like this @alopezari!

zhongruige (approved):
This worked great for me, thanks @alopezari!

PR #378, opened 2025-11-18
Update default performance tests iterations

alopezari:
This PR updates the default number of performance tests iterations from 3 to 7, in order to reduce variability.

### Testing instructions
This PR should be tested together with https://github.com/Automattic/compatibility-dashboard/pull/1574. Testing instructions are defined there.


alopezari:
Merging this as the manager counterpart was tested and approved

PR #379, opened 2025-11-21
Reintroduce UI option

MrJnrman:
### Description

This PR reintroduces the `--ui` option that was available for tests that extended the RunE2ECommand. 

### Testing Instructions
1. Checkout this branch 
2. Using the binary in the root, run any `run:e2e` or `run:activation`  with the `--ui` and confirm that the playwright UI to manually execute tests show up in your browser.


alopezari (approved):
Thanks for adding this @MrJnrman!

LGTM and tested well 💯

zhongruige (approved):
Looks good to me! Thanks for the quick fix here and glad it was only a small change we needed!

PR #388, opened 2025-12-05
Update self-test snapshots

zhongruige:
We've been having self tests failing for a while, likely due to out of date snapshots. This PR updates them.


zhongruige:
Just realized this wasn't on the latest trunk, going to re-run.

PR #387, opened 2025-12-05
Update snapshots

alopezari:
Some tests failed in CI after merging [this PR](https://github.com/woocommerce/qit-cli/pull/384), mostly caused by outdated snapshots (due to the removal of the `iterations` attribute) that weren't flagged in earlier stages of CI pipelines.

This PR updates those snapshots.

PR #389, opened 2025-12-05
Update managed tests snapshots

alopezari:
[This PR](https://github.com/woocommerce/qit-cli/pull/387) didn't update the self-tests snapshots, so we've prepared this one to address it and fix those tests.


alopezari:
Merging this without approval in order to fix the QIT self-tests, since changes are trivial.

zhongruige (approved):
Thanks for getting this in!

PR #392, opened 2026-01-12
Fix failing badges

Luc45:
This PR fixes the failing CI badges in QIT CLI.

**Activation test:** Updated snapshot, normalized lines that comes from WordPress core to avoid the test becoming invalidated on WP Core changes upstream to reduce the maintenance burden of these snapshots.

**Validation test:** Updated snapshot, normalized the list of declarable feature flags to reduce the maintenance burden of these snapshots, replacing it with an assertion that it's a non-empty array instead. Also added a declared compatibility and a declared incompatibility to improve the coverage a bit.

**Custom tests:** Removed it since this has been replaced with Test Packages for a while now.

What's left to do as a follow-up:

- Wrap up and merge the documentation PR: https://github.com/woocommerce/qit-documentation/pull/72
- Add workflow for test package self-tests and add a badge to it
- Once the documentation PR is merged, link to it again from the README


zhongruige (approved):
Changes look good to me, thanks for these updates @Luc45!

PR #398, opened 2026-01-20
Add InfraProviderFactory for provider instantiation

alopezari:
## Summary

- Add `InfraProviderFactory` class for creating infrastructure providers by type
- Initially registers `local` provider (`LocalProvider::class`)
- `cloud` provider will be added in Phase 5 (QIT-882)
- Add unit tests for factory functionality

## Test plan
No need to perform manual testing, a code review is enough at this point.

- [x] `create('local')` returns a `LocalProvider` instance
- [x] `create('unknown')` throws `InvalidArgumentException` with helpful message
- [x] `get_available_types()` returns array of registered type strings
- [x] `register()` allows adding new provider types at runtime
- [x] All unit tests pass (7 tests, 9 assertions)
- [x] PHPCS passes

Closes QIT-869

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/src/Environment/Infra/InfraProviderFactory.php
+++ b/src/src/Environment/Infra/InfraProviderFactory.php
@@ -1,0 +1,60 @@
+<?php
+
+declare( strict_types=1 );
+
+namespace QIT_CLI\Environment\Infra;
+
+/**
+ * Factory for creating infrastructure providers.
+ */
+class InfraProviderFactory {
+	/** @var array<string, class-string<InfraProvider>> */
+	protected array $providers = [
+		'local' => LocalProvider::class,
+		// 'cloud' => CloudProvider::class, // Added in Phase 5 (QIT-882)
+	];
+
+	/**
+	 * Create an infrastructure provider by type.
+	 *
+	 * @param string $type Provider type ('local' or 'cloud').
+	 *
+	 * @return InfraProvider
+	 *
+	 * @throws \InvalidArgumentException If type is unknown.
+	 */
+	public function create( string $type ): InfraProvider {
+		if ( ! isset( $this->providers[ $type ] ) ) {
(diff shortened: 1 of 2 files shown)


src/src/Environment/Infra/InfraProviderFactory.php
```diff
@@ -0,0 +1,60 @@
...
+namespace QIT_CLI\Environment\Infra;
+
+/**
+ * Factory for creating infrastructure providers.
+ */
+class InfraProviderFactory {
+	/** @var array<string, class-string<InfraProvider>> */
+	protected array $providers = [
+		'local' => LocalProvider::class,
+		// 'cloud' => CloudProvider::class, // Added in Phase 5 (QIT-882)
```
zhongruige, line 14: Out of curiosity, was there a reason it was added here but commented out to be included in the follow up? More curious on the process here, rather than just adding it on QIT-882.

alopezari: I believe the agent added it as a placeholder due to documentation purposes. We know from advance the 2 provider types we'll have, and adding this placeholder might explain the initial classes hierarchy (an interface with their implementations and a factory on top of it). Having that placeholder helps understanding why we have this complex hierarchy at this point.

zhongruige: Aligned with what I suspected, interesting seeing this approach.

zhongruige (approved):
LGTM!

src/src/Environment/Infra/InfraProviderFactory.php
alopezari: Yes! I'm taking the opportunity to rely more on AI to build this feature and play the orchestrator and observer roles. It's fun and interesting to see the entire process 😄

PR #406, opened 2026-02-10
Add ecoding to address for 406 errors

zhongruige:
This PR encodes the crtf JSON file that is sent over to avoid 406 errors in workflow runs for larger payloads. See upstream PR for more details.


Luc45 (approved):
Code looks straightforward to me!

PR #409, opened 2026-02-20
Update dir used for group tests

zhongruige:
The group tests have been failing because the directory changed; so this updates the directory used in the GH action to reflect where they currently live.


zhongruige:
Merging to test the changes.

PR #416, opened 2026-03-06
Remove unused bootstrap_log

zhongruige:
This PR removes the unused `bootstrap_log` which wasn't being consumed on the backend or written to.

I don't believe we ever really used this for anything, so just some minor code cleanup.


alopezari (approved):
Nice one spotting this @zhongruige ! LGTM

PR #419, opened 2026-03-13
Sanitize path traversal sequences in CTRF JSON before sending to backend

MrJnrman:
Testing


zhongruige:
@MrJnrman I am thinking this should be good to merge in as part of the CLI release so resolve the conflicts and will do so. Let me know what you think might be the best path forward, since this uses the old file paths (the merge conflict is due to the files moving around).

PR #421, opened 2026-03-17
Bump symfony/process version

zhongruige:
Update symfony/process to patched versions (CVE-2026-24739) in both src/ and integration test dependencies.


zhongruige:
Minor version bump so merging in.

PR #422, opened 2026-03-18
Restore extension set option

zhongruige:
The extension set option was removed, which was causing failures for teams that integrated with it. This restores the option.

### Test instructions

1. Build the binary for this branch
2. Run an activation test, passing in the extension_set option, such as compatibility
3. Verify the test starts and includes all the relevant plugins in the environment

Screenshot 2026-03-18 at 7 36 23 PM


Luc45 (approved):
Code changes looks good to me!

alopezari (approved):
Thanks for working on this @zhongruige! Code LGTM and the activation tests did run the extension set in my local environment 👍

PR #428, opened 2026-03-27
Filesystem-based environment state management

Luc45:
## Summary

- **Replace cache-based EnvironmentMonitor with filesystem scanning** — each environment now writes its full EnvInfo as `env_info.json` in its temp directory, making the filesystem the single source of truth
- **Delete EnvironmentDanglingCleanup** — no longer needed since there's no cache/filesystem divergence to reconcile
- **Add `--all` flag to `env:down`** and `--json` output to `env:list`

## Context

Environment state was tracked in three places that could diverge: a cache blob, per-env JSON files, and temp directories with Docker compose files. When containers were killed externally (system restart, manual `docker compose down`), the cache still showed them as "running" — causing `env:list` to display ghost environments. The `EnvironmentDanglingCleanup` class attempted to reconcile this, but only caught one direction of drift.

With the recent change to non-external Docker volumes/networks, `docker compose down` in the temp directory is fully self-contained, making the cache layer unnecessary.

## Test plan

- [x] `make tests` passes (phpcs, phpstan, phan, phpunit — 192 tests, 495 assertions)
- [x] `env:up` creates environment and writes `env_info.json` to temp directory
- [x] `env:list` shows running environments from filesystem
- [x] `env:list --json` outputs valid JSON
- [x] `env:source` finds most recent environment
- [x] `env:down --all` stops all environments
- [x] `env:down` on environment whose containers were killed externally cleans up correctly
- [x] Old orphan directories without `env_info.json` are invisible to `env:list`


Luc45:
This is a bug fix done as part of using QIT environments in Woo bug blitz.

PR #472, opened 2026-07-09
Require Docker Compose v2 instead of silently falling back to EOL v1

zhongruige:
### Problem

Users running the legacy standalone `docker-compose` (v1) can't start an environment. Any test run ends in:

```
Failed to parse environment JSON. Output:
{"error":"Command failed with non-JSON output","output":"Failed to start the environment. Output:
ERROR: Invalid interpolation format for "qit_env_php_<id>" option in service "services": "${FIXUID:-1000}:${FIXGID:-1000}""}
```

This happens on systems that have Docker Engine installed **without** the `docker-compose-plugin` — `docker compose` (v2) isn't available, so `find_docker_compose()` silently fell back to the standalone `docker-compose` binary. Compose v1 has been EOL since 2023 and cannot parse the environment files QIT generates.

The real failure was the silent fallback: the error surfaced three layers downstream, with nothing connecting it back to "install the Compose plugin."

### Why Compose v1 fails

1. Our generated compose files omit a top-level `version:` key.
2. With no `version` key, v1 resolves the file to **Compose file format 1** (`config.py:195`).
3. On that path v1 interpolates with Python's stdlib `string.Template`, not its own `TemplateWithDefaults` (`interpolation.py:32-36`).
4. Stdlib `Template` rejects any `${VAR-default}` suffix → `Invalid interpolation format`.
5. Under format-1, top-level `services` is read as a *service name* — hence the odd `option "qit_env_php_…" in service "services"` wording.

Docker Compose **v2 parses the same file correctly.** The compose files are valid; this is purely a v1 problem.

### Fix

`find_docker_compose()` no longer silently degrades to v1. If only the legacy binary is present, it throws with install instructions:

```
Docker Compose v2 is required, but only the legacy "docker-compose" (v1) was found.
Docker Compose v1 is end-of-life and cannot run QIT environments.

Install the Compose v2 plugin, then verify with: docker compose version
  Debian/Ubuntu: sudo apt-get install docker-compose-plugin
  RHEL/Fedora:   sudo dnf install docker-compose-plugin
  Other systems: https://docs.docker.com/compose/install/
```

The "nothing found at all" message got the same treatment.

### Why not just patch the compose files?

The obvious-looking fix — changing `${FIXUID:-1000}` to `${FIXUID-1000}` — **does not work**, and was tried and reverted. Compose v1 rejects the suffix whether or not there's a colon (step 4 above). Worse, it's an active regression: under Compose v2 an *empty* `FIXUID` renders `user: ":"` with the single-dash form, while `:-` correctly falls back to `1000`. Empty values are reachable via `QIT_DOCKER_USER` / `QIT_DOCKER_GROUP`.

Adding `version: "2.1"` back to the generators would work, but means carrying a key Compose v2 warns about, forever, to satisfy a dead parser — and still breaks on v1 < 1.11.

Compose v1 doesn't fail in one place, it fails in many. Refusing it once is bounded work; patching symptoms is not.

### ⚠️ Breaking change

Users still on Compose v1 go from "fails confusingly" to "fails clearly, with instructions." That's the intent — v1 never actually worked with these files. Worth a release-notes line.

`Makefile`, `ci/` and `.github/` all use `docker compose`, so **runners are unaffected.**

### Testing

Reproduce by stubbing `PATH` (no v1 install needed):

```bash
mkdir -p /tmp/fakebin

# `docker compose` unavailable; standalone `docker-compose` present → should throw the v1 message
printf '#!/bin/sh\n[ "$1" = "compose" ] && exit 1\nexit 0\n' > /tmp/fakebin/docker
printf '#!/bin/sh\necho "docker-compose version 1.25.0"\n' > /tmp/fakebin/docker-compose
chmod +x /tmp/fakebin/*

PATH="/tmp/fakebin:$PATH" qit env:up
# Expect: "Docker Compose v2 is required..." — not "Failed to parse environment JSON"

# Remove the v1 stub → should throw "Could not find Docker Compose v2."
rm /tmp/fakebin/docker-compose
PATH="/tmp/fakebin:$PATH" qit env:up
```

Regression check with a normal Compose v2 install, `qit env:up` behaves exactly as before.

### DIFF
--- a/src/src/Environment/Docker.php
+++ b/src/src/Environment/Docker.php
@@ -454,34 +454,48 @@ class Docker {
 	}
 
 	/**
+	 * Docker Compose v1 is end-of-life and cannot run QIT environments. The environment files
+	 * we generate omit the top-level "version" key, which v1 resolves to Compose file format 1.
+	 * On that path v1 interpolates with Python's stdlib string.Template, which does not
+	 * understand "${VAR:-default}" (nor "${VAR-default}"), and fails with "Invalid
+	 * interpolation format".
+	 *
+	 * Note that the binary name does not identify the major version: Compose v2 ships both as a
+	 * "docker compose" CLI plugin and as a standalone "docker-compose" binary, so both
+	 * candidates have to be probed for their reported version.
+	 *
 	 * @return array<string> The docker-compose command to use, in a Symfony Process format.
 	 */
 	public function find_docker_compose(): array {
-		$output     = [];
-		$return_var = 0;
+		$candidates = [
+			[ 'docker', 'compose' ],
+			[ 'docker-compose' ],
+		];
 
-		// Docker Compose v2.
-		exec( 'docker compose version >/dev/null 2>&1', $output, $return_var );
-		if ( $return_var === 0 ) {
-			return [ 'docker', 'compose' ];
+		$found_unsupported = null;
+
+		foreach ( $candidates as $candidate ) {
+			$major = $this->get_docker_compose_major_version( $candidate );
+
+			if ( is_null( $major ) ) {
+				continue;
+			}
+
+			if ( $major >= 2 ) {
+				return $candidate;
+			}
+
+			// Installed, but too old. Keep probing: a supported version may exist under the other name.
+			$found_unsupported = implode( ' ', $candidate );
 		}
 
-		/*
-		 * The legacy standalone "docker-compose" (v1) is end-of-life and cannot run QIT
-		 * environments. The environment files we generate omit the top-level "version" key,
-		 * which v1 resolves to Compose file format 1. On that path v1 interpolates with
-		 * Python's stdlib string.Template, which does not understand "${VAR:-default}"
-		 * (nor "${VAR-default}"), and fails with "Invalid interpolation format".
-		 *
-		 * Falling back to it silently produced confusing downstream errors, so refuse it
-		 * explicitly and tell the user how to install v2.
-		 */
-		exec( 'docker-compose version >/dev/null 2>&1', $output, $return_var );
-		if ( $return_var === 0 ) {
+		if ( ! is_null( $found_unsupported ) ) {
 			throw new \RuntimeException(
-				"Docker Compose v2 is required, but only the legacy \"docker-compose\" (v1) was found.\n" .
-				"Docker Compose v1 is end-of-life and cannot run QIT environments.\n\n" .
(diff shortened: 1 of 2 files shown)


src/src/Environment/Docker.php
```diff
@@ -460,18 +460,41 @@ public function find_docker_compose(): array {
...
 		if ( $return_var === 0 ) {
-			return [ 'docker-compose' ];
+			throw new \RuntimeException(
+				"Docker Compose v2 is required, but only the legacy \"docker-compose\" (v1) was found.\n" .
+				"Docker Compose v1 is end-of-life and cannot run QIT environments.\n\n" .
+				"Install the Compose v2 plugin, then verify with: docker compose version\n" .
+				"  Debian/Ubuntu: sudo apt-get install docker-compose-plugin\n" .
+				"  RHEL/Fedora:   sudo dnf install docker-compose-plugin\n" .
+				'  Other systems: https://docs.docker.com/compose/install/'
+			);
```
alopezari, line 502: I think this code rejects every docker compose version, not just v1. Apologies if I'm wrong, haven't tested it, just doing a visual review of the code.

Maybe we'd need to add a condition that ensures the docker compose version we accept/reject?

Additionally, do we want to allow only docker compose v2 or any version higher than v1? According to the title, I assume we want to reject v1 because it was EOL and accept higher ones, right? Wanted to confirm it just in case.

alopezari (commented):
Thanks for working on this @zhongruige!

Added some questions regarding the logic that rejects v1.

zhongruige:
Thanks @alopezari! I'll go into this today, appreciate you checking.

src/src/Environment/Docker.php
zhongruige: Good catch! This needed a version condition (though it wouldn't have rejected every version). On the latter part, the fallback branch treated the docker-compose name as proof of v1, which stopped being true once Compose v2 began shipping as a standalone binary, so anyone on standalone v2 without the plugin would have been hard-failed by the original commit. So, good catch here!

zhongruige: Fixed in e8338da2cde429872fe733e69d65704d73812662

alopezari (approved):
Thanks @zhongruige! LGTM and testing instructions passed

zhongruige:
Thanks for taking the time to help review @alopezari! 🙏

PR #476, opened 2026-07-22
Implement subpackage option for local e2e test runs

daledupreez:
This PR explores the addition of a `subpackage` option to the `qit run:e2e` command that can be used for local e2e test runs to limit execution to one or more subpackages defined for the current local directory. The option can only be used when exactly one local test package is specified, and will error in all other cases.

The gap/use case for this was flagged in [this comment](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5208#issuecomment-5039321745) on https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5208 by @zhongruige -- this PR tries to add support in QIT itself.

I am not actually sure how to test this in my environment, and I put this together using Claude, so we may want to drop the feature.

### DIFF
--- a/src/src/Commands/RunE2ECommand.php
+++ b/src/src/Commands/RunE2ECommand.php
@@ -129,9 +129,6 @@ class RunE2ECommand extends QITCommand {
 			->addOption( 'passthrough_target', null, InputOption::VALUE_OPTIONAL | InputOption::VALUE_IS_ARRAY,
 			'Test packages that should receive passthrough arguments (multiple allowed)', [] )
 
-			/*
-			─────────────── Shared Options (reused from env:up) ───────────────
-			*/
 			/* ─────────────── E2E-specific options ─────────────── */
 			->addOption(
 				'test-package',
@@ -143,10 +140,14 @@ class RunE2ECommand extends QITCommand {
 			->addOption(
 				'subpackage',
 				null,
-				InputOption::VALUE_OPTIONAL | InputOption::VALUE_IS_ARRAY,
+				InputOption::VALUE_REQUIRED | InputOption::VALUE_IS_ARRAY,
 				'Run only the given subpackage(s) of a single local test package; the value(s) must match the subpackage ID(s) in the test package manifest',
 				[]
 			)
+
+			/*
+			─────────────── Shared Options (reused from env:up) ───────────────
+			*/
 			->reuseOption( 'env:up', 'environment' )
 			->reuseOption( 'env:up', 'php_version' )
 			->reuseOption( 'env:up', 'wordpress_version' )
--- a/src/src/Commands/RunE2ECommand.php
+++ b/src/src/Commands/RunE2ECommand.php
@@ -401,7 +401,9 @@ class RunE2ECommand extends QITCommand {
 					if ( ! empty( $requested_subpackages )
 						&& $subpackage_parent_dir !== null
 						&& $pkg_info['source'] === 'local'
-						&& realpath( $pkg_info['path'] ) === realpath( $subpackage_parent_dir ) ) {
+						&& realpath( $pkg_info['path'] ) === realpath( $subpackage_parent_dir )
+						&& false !== realpath( $pkg_info['path'] )
+					) {
 						try {
 							$expanded = $this->expand_subpackages( $pkg_info, $requested_subpackages );
 						} catch ( \InvalidArgumentException $e ) {
--- a/src/src/Commands/RunE2ECommand.php
+++ b/src/src/Commands/RunE2ECommand.php
@@ -220,16 +220,7 @@ class RunE2ECommand extends QITCommand {
 			try {
 				$subpackage_parent_dir = \QIT_CLI\Utils\SubpackageSelector::validate_selection( $requested_subpackages, $candidate_refs );
 			} catch ( \RuntimeException $e ) {
-				if ( $input->getOption( 'json' ) ) {
-					$output->write( json_encode( [
-						'error'   => 'invalid_subpackage',
-						'message' => $e->getMessage(),
-					] ) );
-				} else {
-					$output->writeln( sprintf( '<error>%s</error>', $e->getMessage() ) );
-				}
-
-				return Command::FAILURE;
+				return $this->fail_subpackage_selection( $input, $output, $e->getMessage() );
 			}
 		}
 
@@ -411,25 +402,16 @@ class RunE2ECommand extends QITCommand {
 						&& $subpackage_parent_dir !== null
 						&& $pkg_info['source'] === 'local'
 						&& realpath( $pkg_info['path'] ) === realpath( $subpackage_parent_dir ) ) {
-						$parent_manifest = new \QIT_CLI\PreCommand\Objects\TestPackageManifest( $pkg_info['manifest'] );
-						foreach ( $requested_subpackages as $subpackage_id ) {
-							try {
-								$subpackage_manifest = $parent_manifest->create_subpackage_manifest( $subpackage_id );
-							} catch ( \InvalidArgumentException $e ) {
-								// Defensive: the selection was validated before env:up ran.
-								throw new \RuntimeException( $e->getMessage() );
-							}
+						try {
+							$expanded = $this->expand_subpackages( $pkg_info, $requested_subpackages );
+						} catch ( \InvalidArgumentException $e ) {
+							// This should not happen, as we validate subpackages early on, but this
+							// serves as an additional defensive measure.
+							return $this->fail_subpackage_selection( $input, $output, $e->getMessage() );
+						}
 
-							$test_packages[ $subpackage_id ] = [
-								'manifest'       => $subpackage_manifest,
-								'path'           => $pkg_info['path'],           // Parent host dir (shared).
-								'container_path' => $pkg_info['container_path'], // Parent container path (shared mount).
-								'metadata'       => [
-									'downloaded_path' => $pkg_info['path'],
-									'version'         => 'local',
-									'remote'          => false,
-								],
-							];
+						foreach ( $expanded as $subpackage_id => $entry ) {
+							$test_packages[ $subpackage_id ] = $entry;
 						}
 						continue;
 					}
@@ -452,6 +434,23 @@ class RunE2ECommand extends QITCommand {
 						],
 					];
 				}
+			}
+		}
+
+		// Ensure that when --subpackage is used, every requested subpackage
+		// has an entry in $test_packages. This ensures we only run requested 
+		// subpackages and not the full parent package.
+		if ( ! empty( $requested_subpackages ) ) {
+			$missing_subpackages = array_diff( $requested_subpackages, array_keys( $test_packages ) );
+			if ( ! empty( $missing_subpackages ) ) {
+				return $this->fail_subpackage_selection(
+					$input,
+					$output,
+					sprintf(
+						"Failed to apply the --subpackage selection. The following subpackage(s) were not expanded from the local parent package:\n  - %s\n\nThe full parent package was NOT run, to avoid executing unintended tests.",
+						implode( "\n  - ", $missing_subpackages )
+					)
+				);
 			}
 		}
 
@@ -1194,6 +1193,60 @@ class RunE2ECommand extends QITCommand {
 				}
 			}
 		}
+	}
+
+	/**
+	 * Expand a local parent test package into one entry per selected subpackage.
+	 *
+	 * All entries share the parent's host directory and container mount; only the
+	 * 'run' phase differs per subpackage.
+	 *
+	 * @param array<string,mixed> $pkg_info              Reconstructed parent package info from env_info (needs 'manifest', 'path', 'container_path').
+	 * @param array<string>       $requested_subpackages Subpackage IDs to expand.
+	 *
+	 * @return array<string,array<string,mixed>> Map of subpackage ID => reconstructed test package entry.
+	 * @throws \InvalidArgumentException If a subpackage manifest cannot be synthesized.
+	 */
+	private function expand_subpackages( array $pkg_info, array $requested_subpackages ): array {
+		$parent_manifest = new \QIT_CLI\PreCommand\Objects\TestPackageManifest( $pkg_info['manifest'] );
+		$entries         = [];
+
+		foreach ( $requested_subpackages as $subpackage_id ) {
+			$entries[ $subpackage_id ] = [
+				'manifest'       => $parent_manifest->create_subpackage_manifest( $subpackage_id ),
+				'path'           => $pkg_info['path'],           // Parent host dir (shared).
+				'container_path' => $pkg_info['container_path'], // Parent container path (shared mount).
+				'metadata'       => [
+					'downloaded_path' => $pkg_info['path'],
+					'version'         => 'local',
+					'remote'          => false,
+				],
+			];
+		}
+
+		return $entries;
+	}
+
+	/**
+	 * Emit a --subpackage selection failure using the command's output mode.
+	 *
+	 * @param QITInput        $input The input interface.
+	 * @param OutputInterface $output The output interface.
+	 * @param string          $message The error message.
+	 *
+	 * @return int The exit status.
+	 */
+	private function fail_subpackage_selection( QITInput $input, OutputInterface $output, string $message ): int {
+		if ( $input->getOption( 'json' ) ) {
+			$output->write( json_encode( [
+				'error'   => 'invalid_subpackage',
+				'message' => $message,
+			] ) );
+		} else {
+			$output->writeln( sprintf( '<error>%s</error>', $message ) );
+		}
+
+		return Command::FAILURE;
 	}
 
 
--- a/src/src/Commands/RunE2ECommand.php
+++ b/src/src/Commands/RunE2ECommand.php
@@ -140,6 +140,13 @@ class RunE2ECommand extends QITCommand {
 				'Test packages to include (multiple values allowed)',
 				[]
 			)
+			->addOption(
+				'subpackage',
+				null,
+				InputOption::VALUE_REQUIRED | InputOption::VALUE_IS_ARRAY,
+				'Run only the given subpackage(s) of a single local test package (full ID, e.g. namespace/name; multiple allowed)',
+				[]
+			)
 			->reuseOption( 'env:up', 'environment' )
 			->reuseOption( 'env:up', 'php_version' )
 			->reuseOption( 'env:up', 'wordpress_version' )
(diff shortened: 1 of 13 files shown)


zhongruige:
Ah! This is great @daledupreez! You beat me to looking into this! I'll go through this asap, just wanted to comment so you know it's on my radar!

src/src/PreCommand/Objects/TestPackageManifest.php
```diff
@@ -437,6 +437,83 @@ public function get_subpackage( string $identifier ): ?array {
...
+				// Results paths inherited from parent
+				'results' => $this->test_results,
+			],
+			// Inherit all other configurations from parent
+			'requires'       => $this->requires,
+			'mu_plugins'     => $this->mu_plugins,
+			'envs'           => $this->env_vars,
+			'timeout'        => $this->timeout,
+			'retry'          => $this->retry,
+		];
```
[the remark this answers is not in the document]
daledupreez: Addressed in 1b173f6e440f63c46c214c9062181197385f55b6.

src/src/Commands/RunE2ECommand.php
```diff
@@ -140,6 +140,13 @@ protected function configureMainOptions(): void {
 				'Test packages to include (multiple values allowed)',
 				[]
 			)
+			->addOption(
+				'subpackage',
+				null,
+				InputOption::VALUE_OPTIONAL | InputOption::VALUE_IS_ARRAY,
```
zhongruige: A bare `--subpackage` is returned by Symfony as `[null]`. The later `array_filter()` removes it, so QIT silently runs the complete parent package instead of reporting invalid input.

That is potentially unsafe when subpackages target different regional accounts. Could we require the value and add a regression test?

```suggestion
InputOption::VALUE_REQUIRED | InputOption::VALUE_IS_ARRAY,

**Review by zhongruige — COMMENTED** [review]
I think the chnages here work for the intended use case: selecting a local subpackage replaces the parent package’s `run` phase while inheriting its setup, teardown, requirements, and result configuration.

For example, this produces the expected package:

```sh
qit run:e2e \
  --config=qit.json \
  --profile=main \
  --subpackage=<slug>/<package-name> \
  --env_file=./tests/e2e/qit/qit.au.env
```

I found one behavior that I think should be fixed in this PR, which I left a comment on.

There's work we should do on top of this (I don't think we need to do it on this PR and what is done here is worth having), so I suggest a follow-up that makes subpackage selection part of common package resolution:

* Resolve local and remote subpackages before env:up.
* Attach selection to its parent package so multiple local parents are unambiguous.
* Support selections in profiles and groups.
* Merge environment requirements and report conflicts.
* Give local and Manager-backed subpackages the same resolved-package representation.

Overall otherwise this is looking good!

src/src/Utils/SubpackageSelector.php
```diff
@@ -0,0 +1,113 @@
...
+		// Classify each reference as a local package directory or a remote reference.
+		$local_dirs  = [];
+		$remote_refs = [];
+		foreach ( $package_refs as $ref ) {
+			if ( is_dir( $ref ) && file_exists( rtrim( $ref, '/' ) . '/qit-test.json' ) ) {
+				$local_dirs[] = $ref;
+			} else {
+				$remote_refs[] = $ref;
+			}
+		}
```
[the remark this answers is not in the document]
daledupreez: I addressed this in 2b75a4cf3b04999dc489e40210bc4011429535b7 by using the existing helper methods for packages, and then throwing an exception when `qit-test.json` doesn't exist for the package.

src/src/Commands/RunE2ECommand.php
daledupreez: I updated the option configuration to `VALUE_REQUIRED` in 9b09167e18cec930506f2027c219c8b1028d2596, and added a unit test in dc7249e7a7b77838b5ecc0be2539faab7827a785 and an integration test in 8f860b09309246479eabbe774dabdcbaee2b82f9.

src/tests/unit/SubpackageSelectorTest.php
```diff
@@ -0,0 +1,185 @@
...
+	/**
+	 * Test that a remote package reference is rejected, even alongside a local package.
+	 */
+	public function test_remote_reference_is_rejected(): void {
+		$dir = $this->create_package_dir( $this->get_parent_manifest_data() );
+
+		$this->expectException( \RuntimeException::class );
+		$this->expectExceptionMessage( 'Remote test package references are not allowed when using --subpackage' );
+		$this->expectExceptionMessage( 'woocommerce/e2e:latest' );
+
```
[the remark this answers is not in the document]
daledupreez: Addressed in 219da524431af401b36ee911ab96a15aafa59e67.

src/tests/integration/tests/Commands/RunE2ECommandTest.php
```diff
@@ -271,22 +295,71 @@ public function test_run_e2e_with_local_package_and_custom_plugin(): void {
...
+							'test' => [
+								'phases' => [
+									'setup' => [
+										"echo 'Setting up SUBPACKAGE'",
+									],
+									'run'   => [
+										"npx playwright test $testFile",
+									],
+								],
+							],
```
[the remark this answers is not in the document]
daledupreez: Fixed in 4a0fd9cc116b7df038d7b2b9f02e282784474c04.

src/src/Utils/SubpackageSelector.php
```diff
@@ -0,0 +1,125 @@
...
+		$remote_refs = [];
+		foreach ( $package_refs as $ref ) {
+			if ( PackageReferenceUtils::is_local_reference( $ref ) ) {
+				$local_dir = PackageReferenceUtils::expand_local_path( $ref );
+				if ( ! file_exists( rtrim( $local_dir, '/' ) . '/qit-test.json' ) ) {
+					throw new \RuntimeException(
+						"The local test package directory {$local_dir} does not contain a qit-test.json file.\n" .
+						'Unable to validate the specified subpackages.'
+					);
+				}
```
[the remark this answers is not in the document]
daledupreez: Addressed in b1eb06c3762de193ba2d56744914254983ad9d88.

src/src/PreCommand/Objects/TestPackageManifest.php
```diff
@@ -437,6 +437,77 @@ public function get_subpackage( string $identifier ): ?array {
...
+					]
+				),
+				'results' => $this->test_results,
+			],
+			'requires'       => $this->requires,
+			'mu_plugins'     => $this->mu_plugins,
+			'envs'           => $this->env_vars,
+			'timeout'        => $this->timeout,
+			'retry'          => $this->retry,
+		];
```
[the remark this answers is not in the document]
daledupreez: Addressed in c7723cf28659cb3ee6189527b5f9f2f0302609c1. The code now starts from the `parent->to_array()` data, which should include all fields by default.

src/src/Utils/SubpackageSelector.php
```diff
@@ -0,0 +1,125 @@
...
+
+				$local_dirs[] = $local_dir;
+			} else {
+				$remote_refs[] = $ref;
+			}
+		}
+
```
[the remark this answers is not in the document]
daledupreez: Addressed in bddcfae49974b2fb295cf63924c305e153515862 -- the values should already be canonical, so I've added `array_values( array_unique( ) )` for this check.

src/tests/integration/tests/Commands/RunE2ECommandTest.php
```diff
@@ -271,22 +296,65 @@ public function test_run_e2e_with_local_package_and_custom_plugin(): void {
...
+
+				$manifest['subpackages'] = [
+					$subpackageId => [
+						'description' => 'Subpackage that runs only the example spec',
+						'test' => [
+							'phases' => [
+								// Add custom output to the subpackage's run phase for later verification.
+								'run' => [
+									"echo 'RUNNING_SUBPACKAGE'",
+									"npx playwright test $testFile",
```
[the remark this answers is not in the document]
daledupreez: Fixed in fb89e9dbbad12ade06cdab170f56ccfe0f56df38.

src/src/PreCommand/Objects/TestPackageManifest.php
```diff
@@ -437,6 +439,63 @@ public function get_subpackage( string $identifier ): ?array {
...
+				"Subpackage '{$subpackage_id}' not found in package '{$this->package_id}'."
+			);
+		}
+
+		// Start with parent's complete configuration.
+		$package_data = $this->to_array();
+
+		$package_data['package_id']     = $subpackage_id;
+		$package_data['parent_package'] = $this->package_id;
+		$package_data['subpackages']    = [];
```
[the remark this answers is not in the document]
zhongruige: One extra reason this is legitimately reachable: the Manager resolves a subpackage namespace independently and does not require it to match the parent. A cross-namespace subpackage therefore gets the parent namespace in CTRF/lifecycle metadata here, and `get_package_name()` is always left as the parent name.

There is also a small validation regression compared with the old raw-manifest path: a slash-less subpackage ID used to fail identity construction, while this normalized path silently inherits the parent identity fields. Could we validate/split `$subpackage_id` here and update `namespace` and `package_name` alongside `package_id`?

src/src/PreCommand/Objects/TestPackageManifest.php
```diff
@@ -258,6 +259,7 @@ public function to_array(): array {
 			'parent_package'   => $this->parent_package,
 			'requires_network' => $this->requires_network,
 			'requires_tunnel'  => $this->requires_tunnel,
+			'actions'          => $this->actions,
```
zhongruige, line 273: The `actions` round-trip looks half-done here. Adding it to `to_array()`/`load_from_normalized()` is good, but `jsonSerialize()` still omits it, and `env:up` uses that serializer for the remote-package handoff to `run:e2e`. Local packages pass the raw manifest through, so the rebuilt manifest objects are asymmetric. Runtime action discovery currently rereads the unzipped file, but the manifest serialization contract still drops the field. Could we include `actions` in `jsonSerialize()` and cover both round trips?

src/src/PreCommand/Objects/TestPackageManifest.php
```diff
@@ -437,6 +439,63 @@ public function get_subpackage( string $identifier ): ?array {
...
+		}
+
+		// Start with parent's complete configuration.
+		$package_data = $this->to_array();
+
+		$package_data['package_id']     = $subpackage_id;
+		$package_data['parent_package'] = $this->package_id;
+		$package_data['subpackages']    = [];
+
+		$package_data['phases']['run'] = [];
```
zhongruige: Clearing the run phase here looks dead: the method either throws for a missing subpackage run phase or overwrites this value below. I would drop this assignment, or move the missing-run check above the overrides so the flow reads validate-then-build.

src/src/PreCommand/Objects/TestPackageManifest.php
```diff
@@ -437,6 +439,63 @@ public function get_subpackage( string $identifier ): ?array {
...
+			$package_data['tags'] = $subpackage_config['tags'];
+		}
+
+		if ( isset( $subpackage_config['requires'] ) ) {
+			$package_data['requires'] = array_merge(
+				$this->requires,
+				$subpackage_config['requires']
+			);
+		}
+
```
zhongruige: This line contains a trailing space. `git diff --check` and PHPCS both flag it, so it will need to go before this can pass the style gate.

src/src/Commands/RunE2ECommand.php
```diff
@@ -355,6 +394,28 @@ protected function doExecute( QITInput $input, OutputInterface $output ): int {
...
 				// Check if this package has manifest data (added by env:up)
 				if ( ! empty( $pkg_info['manifest'] ) ) {
+					// When --subpackage is used, expand the single local parent package
+					// into one entry per selected subpackage. All entries share the
+					// parent's host directory and container mount; only the 'run'
+					// phase differs per subpackage (pure subset contract).
+					if ( ! empty( $requested_subpackages )
+						&& $subpackage_parent_dir !== null
+						&& $pkg_info['source'] === 'local'
+						&& realpath( $pkg_info['path'] ) === realpath( $subpackage_parent_dir ) ) {
```
zhongruige: Small edge case: if both paths disappear between the earlier validation and this point, both `realpath()` calls return `false`, and `false === false` matches. Unlikely, but a guard on the resolved package path would make this airtight and avoid expanding the wrong local package.

src/src/Utils/SubpackageSelector.php
```diff
@@ -0,0 +1,126 @@
+<?php
+
+namespace QIT_CLI\Utils;
+
+use QIT_CLI\PreCommand\Configuration\Parser\TestPackageManifestParser;
+use QIT_CLI\Utils\PackageReferenceUtils;
```
zhongruige: `PackageReferenceUtils` is already in the `QIT_CLI\Utils` namespace, so importing it from the same namespace is redundant.

src/src/Utils/SubpackageSelector.php
```diff
@@ -0,0 +1,126 @@
...
+	 * Validate a --subpackage selection against the candidate test packages.
+	 *
+	 * @param array<string> $subpackage_ids Requested subpackage IDs (full IDs, e.g. "namespace/name").
+	 * @param array<string> $package_refs   Candidate test package references, with local paths already expanded to absolute paths.
+	 *
+	 * @return string Absolute path of the single local parent package directory.
+	 * @throws \RuntimeException On any violation.
+	 */
+	public static function validate_selection( array $subpackage_ids, array $package_refs ): string {
+		$subpackage_ids = array_values( array_unique( array_filter( $subpackage_ids ) ) );
```
zhongruige: The requested IDs are already normalized in `RunE2ECommand` before this method is called. Repeating the `array_filter`/`array_unique` here is harmless, but it makes ownership of normalization a little unclear. I would keep it at the command boundary, since that normalized array is used again for expansion, and remove this copy.

zhongruige (commented):
A few things I noticed while tracing the local subpackage path through manifest serialization and execution.

src/src/PreCommand/Objects/TestPackageManifest.php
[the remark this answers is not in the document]
daledupreez: Addressed in 2c1bedab3e3fe0b20e9101a540c09321641fee86, which implements both validation for the subpackage ID and recomputation of the namespace and package name for the TestPackageManifest based on the subpackage ID.

src/src/Utils/SubpackageSelector.php
```diff
@@ -0,0 +1,126 @@
...
+		}
+
+		if ( ! empty( $remote_refs ) ) {
+			throw new \RuntimeException(
+				"The --subpackage option can only be used with a single local test package.\n" .
+				"Remote test package references are not allowed when using --subpackage:\n" .
+				'  - ' . implode( "\n  - ", $remote_refs ) . "\n" .
+				"\n" .
+				"To run a remote subpackage, reference it directly, e.g.:\n" .
+				'  --test-package namespace/subpackage:version'
```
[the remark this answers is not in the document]
daledupreez: Applied in 9aa12f91e3f8ef063aea19c306a5c494ed36e96d.

src/src/PreCommand/Objects/TestPackageManifest.php
daledupreez: I added actions to the JSON serialisation and implemented full round trip unit tests in e6dcda5a3d1a9b2a1f917a1680828c0bf1d54662.

src/src/PreCommand/Objects/TestPackageManifest.php
daledupreez: Removed in 1adb74c150b4b78bcd4057d0000c05d6301f7071.

src/src/PreCommand/Objects/TestPackageManifest.php
daledupreez: This got picked up during one of the other commits. 😁

src/src/Commands/RunE2ECommand.php
daledupreez: Added in 08dcfe03dd17647fee1ba7157a6cdc0897c7a250.

src/src/Utils/SubpackageSelector.php
daledupreez: I removed the `use` in a2413d563e83cb1612812e06f19703373db7f06c.

src/src/Utils/SubpackageSelector.php
daledupreez: Removed in e5dc1aec4d938cbe0e2dc1c1e209576305f5a2ec.

zhongruige:
@daledupreez just wanted to check if you were ready for me to take another look at this. Thanks!

daledupreez:
@zhongruige, yes, I would love for you to take another look!

src/src/Commands/RunE2ECommand.php
```diff
@@ -185,6 +193,37 @@ protected function doExecute( QITInput $input, OutputInterface $output ): int {
 			return self::FAILURE;
 		}
 
+		/* ─ Validate --subpackage selection (fail fast, before any expensive work) ─ */
+		$requested_subpackages = array_values( array_unique( array_filter( (array) $input->getOption( 'subpackage' ) ) ) );
```
zhongruige, line 197: [P1] Please reject explicitly empty selections here instead of filtering them away. Symfony parses `--subpackage=` as `['']`; this `array_filter()` turns that into `[]`, skips the entire selection branch, and the normal path runs the full parent package. This is especially easy to hit with `--subpackage="$TARGET"` when an environment variable is unset, and it contradicts the later “full parent package was NOT run” safety guarantee. Please check raw option presence and fail on any empty or whitespace-only value, with coverage for `--subpackage=`.

src/src/Commands/RunE2ECommand.php
```diff
@@ -1112,6 +1197,60 @@ private function validate_subpackage_versions( array $test_packages ): void {
...
+	 * @return array<string,array<string,mixed>> Map of subpackage ID => reconstructed test package entry.
+	 * @throws \InvalidArgumentException If a subpackage manifest cannot be synthesized.
+	 */
+	private function expand_subpackages( array $pkg_info, array $requested_subpackages ): array {
+		$parent_manifest = new \QIT_CLI\PreCommand\Objects\TestPackageManifest( $pkg_info['manifest'] );
+		$entries         = [];
+
+		foreach ( $requested_subpackages as $subpackage_id ) {
+			$entries[ $subpackage_id ] = [
+				'manifest'       => $parent_manifest->create_subpackage_manifest( $subpackage_id ),
```
zhongruige, line 1218: [P1] This synthesis happens only after `run_environment()` has already provisioned the environment using the parent directory. Any allowed `subpackages.*.requires` override for plugins, themes, network, or tunnel is therefore ignored for a local selection. The new manifest unit fixture even has a cart subpackage requiring `woocommerce-gateway-stripe`, but that dependency will not be installed unless the parent also requires it. Please derive and merge the selected requirements before `env:up`, or teach `env:up` about the selection, so local behavior matches remote subpackage semantics.

src/src/Commands/RunE2ECommand.php
```diff
@@ -1112,6 +1197,60 @@ private function validate_subpackage_versions( array $test_packages ): void {
...
+				'path'           => $pkg_info['path'],           // Parent host dir (shared).
+				'container_path' => $pkg_info['container_path'], // Parent container path (shared mount).
```
zhongruige, line 1220: [P2] Multiple selected siblings each get a metadata entry pointing at this same parent path. `PackagePhaseRunner::generate_actions_manifest()` scans every metadata entry and reads the raw parent manifest, so the same action provider/path is appended once per selected sibling. `qit.actions()` then returns duplicate callables and consumers can execute the same action twice. Please deduplicate by provider/action path (or contribute runtime metadata once per physical package) and add coverage for multiple selected subpackages whose parent declares actions.

src/src/Utils/SubpackageSelector.php
```diff
@@ -0,0 +1,146 @@
...
+		}
+
+		if ( empty( $local_dirs ) ) {
+			throw new \RuntimeException(
+				"The --subpackage option requires a local test package.\n" .
+				'Provide one with --test-package <dir>, via qit.json, or run from a directory containing qit-test.json.'
+			);
+		}
+
+		$normalized_local_dirs = array_values( array_unique( $local_dirs ) );
```
zhongruige, line 71: [P2] `array_unique()` compares path spellings rather than directory identities. `expand_local_path()` deliberately returns absolute paths unchanged, so `/path/package` and `/path/package/.`—or a directory and a symlink alias—are counted as two local parents. I reproduced the error with the fixture path plus `/.`. Please canonicalize each validated directory with `realpath()` before deduplicating and counting it.

zhongruige (commented):
Thanks for addressing the earlier review threads. I found four additional edge cases in the current head and left inline comments with suggested fixes and coverage.

src/src/Commands/RunE2ECommand.php
daledupreez: This should now be addressed via 865d7c2401fad41dc4fc44920d02ea4a8ad2e9ba.

src/src/Commands/RunE2ECommand.php
daledupreez: This should be addressed by edab7013b845067c90422fb3e8f77f59ae27713b.

src/src/Commands/RunE2ECommand.php
daledupreez: This should be addressed by a4488eb9179e208f62e358ae15662f2e020d907f.

src/src/Utils/SubpackageSelector.php
daledupreez: Addressed in 9d93f3f587b718bd6f20ecde65e12379ae78de31.

PR #492, opened 2026-09-08
Pick the nightly package for WooCommerce ahead of every published line

alopezari:
## What changed

`run:activation` had one answer for a WooCommerce version no published package covers: `latest`. Now it has two. A version ahead of every published line takes `woocommerce/activation:nightly`; everything else keeps `latest`.

```
--woo=nightly  → No test package covers WooCommerce nightly. Using woocommerce/activation:nightly instead.
--woo=11.1.0   → Using test package woocommerce/activation:11.1 for WooCommerce 11.1.0.
```

## Why

`latest` follows the newest published stable line, so it was the wrong suite for the case that comes round every cycle: a nightly, or the next line's prerelease, running trunk's markup rather than the last release's.

That is not hypothetical. `run:activation woocommerce-bookings --woo=nightly` failed on WooCommerce 11.2.0-dev, because trunk changed the add-products modal and `latest` still carried the 11.1 specs.

## How it works

Two conditions, not one. The version has to be ahead of every published line, compared on `major.minor` so a prerelease counts as the line it belongs to: 11.2.0-rc.1 is ahead while 11.1 is the newest published, and stops being ahead the moment 11.2 is published. It also has to be one WooCommerce has not released — `nightly`, a prerelease, a `-dev` build.

The second condition is what keeps the default run out of this. A line can go GA before its package is published, `stable` resolves to it, and a plain `run:activation` with no `--woo` would otherwise take trunk's suite against a released store for the length of that gap.

`nightly_test_package()` is concrete and returns null rather than abstract. `run:woo-e2e` shares this trait and has no nightly tag published, so an abstract method would have pointed its runs at a package that does not exist. A command opts in by overriding; one that has not returns exactly what it did before.

## Validation

- `make phpunit` — 487 tests, 1406 assertions, green. Five new cases: ahead of every line, the `nightly` channel, behind the newest line, no nightly tag published, and the line ceasing to be ahead once published.
- `make phpcs` clean on the changed files; two pre-existing warnings elsewhere are untouched.
- Ran both commands above against a live Manager, on a branch cut from current `trunk`.

## Related

Follows #490. The Manager applies the same rule for the runs it creates itself: Automattic/compatibility-dashboard#1816.

Merge after Automattic/compatibility-dashboard#1816. That one publishes `woocommerce/activation:nightly` and advertises it in sync data; this reads the tag from there, so against a Manager without it every version keeps `latest` rather than chasing a package that is not published.


alopezari:
Merging this PR since the CI failure is not related to it and the checks passed locally. I'll address the CI issues on a separate task.

